PR #4, opened 2017-12-10
Added localizedPrice, title, description...

KjellConnelly:
Made it a bit easier to work cross platform between iOS and Android and added localizedPrice so you can display the price of an IAP to the user without having to deal with currency symbols and formatting.


KjellConnelly:
Added better error handling (purchase failure) for iOS and updated readme

JJMoon:
Thank you for your effort.

PR #26, opened 2018-02-11
Updating Readme

KjellConnelly:
Just fixing some grammar and examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,38 +118,42 @@ const itemSkus = {
 Next, call the prepare function (ios it's not needed, but android it is. No need to check platform though since nothing will happen in ios:
 
 ```javascript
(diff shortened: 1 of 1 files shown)


dooboolab:
Um why did you delete all the readme below
`#### Each item is a JavaScript object containing these keys:`
??

KjellConnelly:
Take a look at the readme. At some point, a lot of the file got duplicate parts, as if someone copy pasted a large portion of it within itself. If you look in my branch’s readme, you will see it still has the same information, but removes the duplicates. You could double check to make sure I didn’t delete anything important though. Looking at just the changes is a bit confusing, so looking at them side by side may be easier.

dooboolab:
Ah ok! Thanks for your verification.

PR #33, opened 2018-02-21
Aligned format of purchase-response

Froelund:


dooboolab:
Um.. this should be considered before merge because now the result is seperated with "data" and "signature" which will be handled differently from ios.

Froelund:
Well, the formats are already different from each other.

dooboolab:
What will be the validation string for iap? It will just be `data` isn't it? Just to make sure how to make ios and android reliable to each other.

dooboolab:
I will merge this since this is needed. But have to notice users that result string is behaving differently to verify the purchase in their backend.
Will have to see how ios will work with this one too.

PR #39, opened 2018-02-25
WIP: Improve typings, with JSDoc

cometkim:
- Separated platform specified types
- Added JSDoc decorators
- Modified some android specified methods to return null if the Platform is not matched

WIP - Some JSDoc is incomplete because I do not fully understand its behaviors, need help to make a meaningful description.


dooboolab:
Thank you!

Issue #44, opened 2018-03-01
React pod has been deprecated


rikur:
```
Installing RNIap (1.0.0)
Installing React (0.11.0)
..
[!] React has been deprecated
```

I had to remove some other libs that pull in the deprecated React pod as it causes weird errors sometimes. Could this be fixed?

rikur:
Also documentation doesn't mention that you have to:
```
cd ios
pod install
```

after linking the library to pull in the pod.

rikur:
And... I can't use this library:

```Failed to load bundle with error:(Ambiguous resolution: module './index.js' tries to require 'react-native', but there are several files providing this module. You can delete or fix them: 

/Users/rikur/workspace/listberry/ios/Pods/React/package.json
...
```

New React Native and Cocoapods don't work if you pull in the React pod.

rikur:
Tried to fix this in my fork, but ttill getting `TypeError: Cannot read property 'fetchProducts' of undefined`.

rikur:
`react-native link react-native-iap` doesn't seem to work. My forked version and manual linking solved my problems.

dooboolab:
I've just merged this and published to npm `0.2.15`. We'll see how it goes. Thanks.

rikur:
Thanks and wow.. I would have kept it in master for a bit before releasing.. I didn't test every version of React Native/Cocoapods, only the latest.

PR #45, opened 2018-03-01
Remove React dependency -- it breaks builds

rikur:


rikur:
#44

dooboolab:
Ok~ I will merge this and lets see how it works.

rikur:
My bad, I will read the doc.

Issue #55, opened 2018-03-05
BAD CONSUMPTION FUNCTIONALITY!!


lwansbrough:
`refreshAllItems` is very poorly named!! The implementation uses `fetchHistory` on iOS which is non-mutative operation, and uses `consumeItem` on Android, which destroys purchase history! This is either very bad naming, or very poor purchase flow behaviour! I'll nominate it for both!

I can't think of a single case where I'd want to immediately consume all purchased items. What I would expect is that `refreshAllItems` re-synchronizes the device purchase history with Google Play and the App Store (ie. `getPurchases`) -- this should absolutely be changed so more people don't make this mistake.

I even checked the source code to make sure I had the right idea about what was going on, but I got as far as `getPurchases` on the Android side and stopped reading. I didn't notice this module was automatically consuming those purchases!!

Here's my suggestion for a fix: rename `refreshAllItems` to `refreshPurchases`, and ONLY fetch the user's purchase history. Do NOT consume items automatically. Consuming behaviour is it's own flow which should be managed by the application developer, ie explicitly calling `consumeItem(...)` or something similar.

lwansbrough:
https://developer.android.com/google/play/billing/billing_integrate.html

> You typically don't want to implement consumption for in-app products that are purchased once and provide a permanent effect (for example, a premium upgrade).

lwansbrough:
Here's what you need instead of `getPurchases`: you need `getPurchaseHistory`. https://developer.android.com/google/play/billing/billing_reference.html#getPurchaseHistory

dooboolab:
@lwansbrough Hi! thank you for really important testing. As you can see we are out of time just focusing on this module. However, I am trying hard to maintain this so others can use `in-app-purchase` functionality easy as possible. Thank you for the link, I've found out that they implemented new api since `2017.05` for `getPurchaseHistory`. I've just changed method name `refreshAllItems` to `fetchHistory`, I've found out their characteristic was totally different since ios control the purchase process itself. Also if you look in the readme, there is `getPurchasedItemsAndroid` method for android which query for purchased items in android.

Could you please try this [commit](https://github.com/dooboolab/react-native-iap/commit/b8d3002e072953262a89bb879f0d70c9386c2213) which I've just updated? Since I may take some time setting the testing environment, I wish I can hear from you if this is what you want.

Thank you again for the help. If you are fine with this commit, I'll publish this to `0.2.17`.

lwansbrough:
[...]

PR #58, opened 2018-03-06
Major refactoring of code base, including core concepts

lwansbrough:
This started with issue #55 and has evolved a bit to create an easier to understand IAP module. My goal was to abstract away the differences and complexities between platforms while still providing the core benefits of each. This includes changing our common terminology surrounding items, purchases, consumption, and restoration.

Very much open to suggestions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,7 @@
 Please follow the Coding conventions as much as possible when contributing your code.
 * The indent tab is two spaces.
 * The class declaration and the `{}` in curly brackets such as function, if, foreach, for, and while should be in the following format. Also if you installed eslint in vscode or in your code editor, it will help you with linting.
+	* `{` should be placed in same line and `}` should be placed in next line.
 ```
 for (let i = 0; i < 10; i++) {
   ...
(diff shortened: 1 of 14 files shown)


lwansbrough:
I should mention this has also been largely untested. I'm going to be testing soon though.

dooboolab:
Thank you. I will go over it the day after tomorrow since I have some conference tomorrow. 
By the way, could you give pull request to `dev` branch? You can refer to CONTRIBUTING.md.

Froelund:
Nice work 👍 Do you need any help?

lwansbrough:
@Froelund We deployed these changes to production yesterday and everything seems to be working (though we don't use the consumption code.) My suggestion is that this is okay to merge, but should be tested as much as possible.

rikur:
wow, great job! Will test this out.

rikur:
Other than that, seems to work great! If I'm not mistaken, the responses from purchase/restore are now more streamlined and easier to use, too.

lwansbrough:
@riku Thanks for testing! It’s definitely possible we have some bugs on iOS. I said we deployed to prod but I really meant we deployed the Android side of it to prod, still using the old version on iOS. I’ll look into the issues you found.

lwansbrough:
@rikur Also would be helpful to know what version of iOS you’re on. Subscription type checking only works on iOS 10+ I believe.

rikur:
@lwansbrough Running on iOS 11.x, deployment target iOS 9.0

PR #74, opened 2018-03-26
Default itemType & currencyCode to empty string

mars-lan:
As inserting nil into NSDictionary is not allowed


dooboolab:
Thanks.

PR #80, opened 2018-03-27
Default itemType & currencyCode to empty string

mars-lan:
As inserting nil into NSDictionary is not allowed.


mars-lan:
Trying this once again for merging into dev branch

PR #95, opened 2018-04-04
Fix the price conversion to also show the fractional part

neilco:
This pull request fixes an issue where converting the subscription price to a string was truncating the fractional part of the price.


dooboolab:
Please do not make mistake for PR on `dev` branch.
By the way, Thank you for your work.

PR #103, opened 2018-04-09
fix missing char in RNIapModule.java

kasterlod:
fix missing char in RNIapModule.java which causing build/run error on android


dooboolab:
Please commit to `dev` branch instead `master` refering to Contributing guide.

PR #105, opened 2018-04-09
Hotfix/0.3.2.1 - fix missing char in RNIapModule.java

kasterlod:
fix missing char in RNIapModule.java which causing build/run error on android


lwansbrough:
Would be good to get this into a release as soon as possible. This will cause failed purchases on some devices (we've seen it 20 times in the past 2 days.)

dooboolab:
@lwansbrough I’ve already published this to 0.3.3 did you try this one?

lwansbrough:
I see that now, thanks!

PR #108, opened 2018-04-10
fix the bugs about different function name defined in index.js and index.d.ts.

a1onexiang2:
I tried to call `RNIap.consumePurchase(token)` in my React-Native App after calling `RNIap.buyProduct(sku)` and getting product information, but it failed.
After reading the source code, I found the function name is defined as `consumePurchase` in `index.d.ts` but `consumeProduct` in `index.js`, what means `consumePurchase` cannot do anything but throw an exception.
So I just change the function name from `consumeProduct` to `consumePurchase` in `index.js` to satisfy the doc and that's it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -90,7 +90,7 @@ export const buyProduct = (sku) => Platform.select({
  * @param {string} token The product's token (on Android)
  * @returns {Promise}
  */
-export const consumeProduct = (token) => Platform.select({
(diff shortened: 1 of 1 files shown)


dooboolab:
Yes. That's a mistake. Thank you~

a1onexiang2:
@dooboolab The first time I tried to commit a PR. Sorry for that.

PR #111, opened 2018-04-10
Fix null reference crash when initiating billing flow on Android

neilco:
This pull request fixes a crash caused by a null reference exception when initiating the billing flow on Android.


neilco:
Closing since this seems to have been fixed in `master`. Could you please update `dev`? It's currently 5 commits behind `master`.

PR #132, opened 2018-04-19
Pass error code when unexpected error occurs.

benevbright:
Knowing error code helps to resolve the problem. 
NOTE: Error code from iOS can be negative value.


dooboolab:
@christophercotton Thank you for the fix. I've published to `0.3.14`. Also follow the readme that version that is not mentioned in `changelog` may occur some bug because I've not tested that yet.. Sorry.

christophercotton:
How does it get pushed to NPM? Since I just did an install, and as a first time user it made me wonder what had happened. Is it possible to keep those out of NPM until they are tested?

PR #186, opened 2018-06-08
Handle network error

wasedaigo:
When I tested with Airplane mode, getProduct never resolved/rejected the promise. Fix this by handling the delegate properly.


dooboolab:
Thank you for the `PR`. Have you tested this? If so will publish to new npm version.

PR #187, opened 2018-06-09
Revert change in 530e433 that broke CocoaPod-based install on ios

mars-lan:


dooboolab:
@mars-lan Thank you for the fix. Also, we no longer have `dev` branch now. So, next time you can `PR` directly to master.

PR #188, opened 2018-06-10
Android subscription upgrade/downgrade

### LINKED ISSUE #185: Subscription upgrade in Android
1.1.0

Android

### Expected behavior
Be able to upgrade subscription

### Actual behavior
There is no way to upgrade subscription in Android

Real Device

See API

### Description

Hello!

In iOS upgrading is handled by the AppStore when buying **Yearly** while having **Monthly**. However as I mentioned in #171 there is a special way in Android to upgrade subscription:
>However, when upgrading or downgrading, you pass the product IDs for the current subscription and the future (upgraded or downgraded) subscription to the BillingFlowParams object using the addOldSku() method.

https://developer.android.com/google/play/billing/billing_subscriptions#Allow-upgrade

I think it's an important feature to have in IAP.

vshab:
Hello!

Add optional `oldSku` parameter for `buySubscription` on **Android** to support subscription upgrade/downgrade.

Fix #185

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Lastly, this module also supports types for typescript users from `0.2.5`.
 | getSubscriptions | `string[]` Subscription IDs/skus | `Promise<Subscription[]>` | Get a list of subscriptions. Note: On iOS  this method has the same output as `getProducts`. Because iOS does not differentiate between IAP products and subscriptions.  |
 | getPurchaseHistory | | `Promise<Purchase[]>` | Gets an invetory of purchases made by the user regardless of consumption status (where possible) |
 | getAvailablePurchases | | `Promise<Purchase[]>` | Get all purchases made by the user (either non-consumable, or haven't been consumed yet)
-| buySubscription | `string` Subscription ID/sku | `Promise<Purchase>` | Create (buy) a subscription to a sku |
+| buySubscription | `string` Subscription ID/sku, `string` Old Subscription ID/sku (on Android) | `Promise<Purchase>` | Create (buy) a subscription to a sku. For upgrading/downgrading subscription on Android pass second parameter with current subscription ID, on iOS this is handled automatically by store. |
 | buyProduct | `string` Product ID/sku | `Promise<Purchase>` | Buy a product |
 | buyProductWithoutFinishTransaction | `string` Product ID/sku | `Promise<Purchase>` | Buy a product without finish transaction call (iOS only) |
 | finishTransaction | `void` | `void` | Send finishTransaction call to Apple IAP server. Call this function after receipt validation process |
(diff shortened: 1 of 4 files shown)


vshab:
@dooboolab this change shouldn't change anything for exciting apps like in 99% I would say. The oldSku parameter is optional, so existing code should just work. I think that something is possible if somebody for some reason passing extra parameter in buySubscription method which doesn't make any sense now :)
I just didn't want to create more methods which make sense only for only one platform in some specific case because it leads to more sophisticated API.
However if you think that this should be done as an another method I can do it, no problem :)

vshab:
@dooboolab 

>Below are two different methods in js isn't it? Correct me if I am wrong.

Not exactly. When calling you can pass as many arguments to function as you want, the difference is how to access them inside of a function. Lest look at two functions for example:
```js
function foo(a) {
  console.log(a, arguments[1]);
}

function bar(b, c) {
  console.log(b);
}
```

We can call first one with two arguments:
```js
foo(1, 2);
```
and get `1 2`

And call second one with one argument:
```js
bar(3);
```
and get `1`

By default function arguments are set to `undefined`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
>In JavaScript, parameters of functions default to undefined.

So basically in JS function can accept any number of arguments and naming them in definition only helps accessing. If there are less arguments in call than in definition then unspecified arguments will be `undefined`.  That's what happening when old code will call `buySubscription` - `oldSku` by default would be `undefined` and everything should work as before. However new code can set `oldSku` and get new functionality.

>Also, to be optional in ts, you should put ? in index.d.ts

You are right here! I'm not very good in TypeScript :(

vshab:
@dooboolab i updated TypeScript definition.

dooboolab:
@vshab Thank you for your specific explanation. I was confused with the `java`. Have you tested this with non-optional and optional function call? Because, I think `public void buyItemByType(String type, String sku, String oldSku, Promise promise)` and `public void buyItemByType(String type, String sku, Promise promise)` this will be different method in java.

vshab:
@dooboolab yep, I've tested the code.
Here is the call **without** second argument: 


And here is **with**:

dooboolab:
Ok. I will merge this and publish to newer version of npm which will be `1.1.4`. Let's see what happens.

vshab:
@dooboolab 🎉 I hope we won't break anything :)
Thanks for quick responses!

dooboolab:
@vshab Thanks for your new feature. Also, there was a break in your commitment so I quickly fixed this in `react-native-iap@1.1.6'. This is just to inform you if there can be any improvement with my fix. Thank you for your work again.

vshab:
@dooboolab oh.. that's unfortunate.  But good job fixing it! :)
I'll look if there anything else can be done.

PR #192, opened 2018-06-14
add endConnection method to default export

chris-divvito:
This method is missing from the default export currently, meaning you have to import it separately via a barrel brace or face exceptions that it is not a function.


dooboolab:
Thank you

Issue #195, opened 2018-06-20
Return an JSON parse error


wasedaigo:
I got a report from a customer that she does not seem to be able to restore her items correctly. I am suspecting something is failing during the JSON parse of BillingResponse. (for example, if orderID is missing, this would fail) I want to catch this error and throw it to Crashlytics.

Another approach is instead of returning an error, keep an empty field empty instead of relying on GetString() to raise an error. Basically, we should check each field to see whether it exists. This approach is safer and give more control to devs.

What do you think?

https://github.com/dooboolab/react-native-iap/pull/196

dooboolab:
Merged just now.

PR #196, opened 2018-06-20
[Do not merge yet] Return an error when it fails to parse JSON

wasedaigo:
What do you think? 
The downside of this is that it fails the entire retrieval of items in case there is any "bad" transaction exists... 

I also use optString for "orderId" as it is known to be empty for test order before 2018... But perhaps I should just fail that case as well?

https://github.com/dooboolab/react-native-iap/issues/195


wasedaigo:
This change is too risky in a way one "bad" billing response breaks all..

dooboolab:
@wasedaigo Please remind us when you think this is ready to merge.

wasedaigo:
I am waiting for feedbacks

dooboolab:
@wasedaigo Looks great. I think we can merge this. What do you think?

PR #213, opened 2018-07-24
Update build.gradle

idlework:
Add defaults fallback when rootProject.ext does not contain sdk- and compile-versions.

the parameters `compileSdkVersion`, `buildToolsVersion`, `minSdkVersion` and `targetSdkVersion` were not available in the Root project of my project. Adding a standard fallback to the `build.gradle` ensures that the react-native-iap package can always compile for everyone.


dooboolab:
Thanks

martnst:
Hey, I am wondering why it needs min API version 26. Looking at the current distribution shares that's apparently cutting of 90% of the market.

martnst:
Okay, I double checked the changes… it seems that is is now supposed to pull minSdkVersion and so on from the project's config. Unfortunately, however, it does not work for me. I guess it related to #215 

FYI: I am staying at version 1.2.4 for now, which works fine fore me.

mvanroon:
looks like @idlework made a tiny mistake here

fixed in #217

PR #214, opened 2018-07-24
add version code

marshallshen:
The lastest package gave the following error when compiled:

```
* Where:
Build file '/Users/mshen/Projects/native/maestro/node_modules/react-native-iap/android/build.gradle' line: 29

* What went wrong:
A problem occurred evaluating project ':react-native-iap'.
> Cannot invoke method versionCode() on null object

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```


dooboolab:
There wasn't a `enter` key which caused this problem. Thanks.

PR #215, opened 2018-07-25
Revert the default min SDK version back to 16

mars-lan:
As this was the case prior to bf7da273dd2715b9f30242edc7727102e3559902


dooboolab:
@mars-lan Why is this necessary? Today, minSdk that could be published to playstore is 26 i guess. Isn't it?

dooboolab:
Do you think minSdk 21 will be ok?

hariks789:
minimum required targetSDK is 26 not minSDK

dooboolab:
@hariks789 yeah. I realized later. Thanks. Merged this PR already.

hariks789:
@dooboolab thanks 👍

PR #217, opened 2018-07-26
fix minSdkVersion

mvanroon:


dooboolab:
@mvanroon Thanks. This change released to `1.3.4`.

PR #235, opened 2018-08-15
TypeScript Improvements

LinusU:
Recently started using the library in a TypeScript project and noticed a few things that could be improved. Thanks for a great library!

Here are some examples of the improvements seen from VS Code:

screen shot 2018-08-15 at 12 23 43

screen shot 2018-08-15 at 12 30 29

I haven't had time to type up the response for `validateReceiptAndroid` yet, but hopefully there will be another pull request soon™ 🚀

### DIFF
--- a/apple.d.ts
+++ b/apple.d.ts
@@ -1,0 +1,347 @@
+export enum ReceiptValidationStatus {
+  /** The receipt validated successfully. */
+  SUCCESS = 0,
+  /** The App Store could not read the JSON object you provided. */
+  INVALID_JSON = 21000,
+  /** The data in the receipt-data property was malformed or missing. */
+  INVALID_RECEIPT_DATA = 21002,
+  /** The receipt could not be authenticated. */
+  COULT_NOT_AUTHENTICATE = 21003,
+  /** The shared secret you provided does not match the shared secret on file for your account. */
+  INVALID_SECRET = 21004,
+  /** The receipt server is not currently available. */
+  UNAVAILABLE = 21005,
+  /** This receipt is valid but the subscription has expired. When this status code is returned to your server, the receipt data is also decoded and returned as part of the response.
+    Only returned for iOS 6 style transaction receipts for auto-renewable subscriptions. */
+  EXPIRED_SUBSCRIPTION = 21006,
+  /** This receipt is from the test environment, but it was sent to the production environment for verification. Send it to the test environment instead. */
+  TEST_RECEIPT = 21007,
+  /** This receipt is from the production environment, but it was sent to the test environment for verification. Send it to the production environment instead. */
+  PROD_RECEIPT = 21008,
+  /** This receipt could not be authorized. Treat this the same as if a purchase was never made. */
+  COULD_NOT_AUTHORIZE = 21010,
+
+  /** Internal data access error. */ INTERNAL_ERROR_00 = 21100, /** Internal data access error. */ INTERNAL_ERROR_01 = 21101, /** Internal data access error. */ INTERNAL_ERROR_02 = 21102, /** Internal data access error. */ INTERNAL_ERROR_03 = 21103, /** Internal data access error. */ INTERNAL_ERROR_04 = 21104, /** Internal data access error. */ INTERNAL_ERROR_05 = 21105, /** Internal data access error. */ INTERNAL_ERROR_06 = 21106, /** Internal data access error. */ INTERNAL_ERROR_07 = 21107, /** Internal data access error. */ INTERNAL_ERROR_08 = 21108, /** Internal data access error. */ INTERNAL_ERROR_09 = 21109,
+  /** Internal data access error. */ INTERNAL_ERROR_10 = 21110, /** Internal data access error. */ INTERNAL_ERROR_11 = 21111, /** Internal data access error. */ INTERNAL_ERROR_12 = 21112, /** Internal data access error. */ INTERNAL_ERROR_13 = 21113, /** Internal data access error. */ INTERNAL_ERROR_14 = 21114, /** Internal data access error. */ INTERNAL_ERROR_15 = 21115, /** Internal data access error. */ INTERNAL_ERROR_16 = 21116, /** Internal data access error. */ INTERNAL_ERROR_17 = 21117, /** Internal data access error. */ INTERNAL_ERROR_18 = 21118, /** Internal data access error. */ INTERNAL_ERROR_19 = 21119,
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1,31 +1,29 @@
...
+
+interface Common {
+  title: string
+  description: string
+  price: string
+  currency: string
+  localizedPrice: string
 }
 
-export const SkuTypes: SkuTypes
```
LinusU: I'm removing this because it wasn't actually exported from `index.js`, so trying to use it would just result in `undefined` at runtime.

I'm planning on fixing this by normalizing the `type` property so that it's the same on Android and iOS, expect a PR for that soon™ 😎

LinusU:
I removed `RNVersion` from here as well, my reasoning is that passing in `undefined` will actually work since `undefined < 54` evaluates to `false`. Anyhow, when we merge #238 that parameter will be gone forever 🎉

apple.d.ts
```diff
@@ -0,0 +1,347 @@
...
+export interface ReceiptValidationRequest {
+  /** The base64 encoded receipt data. */
+  'receipt-data': string
+  /** *Only used for receipts that contain auto-renewable subscriptions.* Your app’s shared secret (a hexadecimal string). */
+  password?: string
+  /** *Only used for iOS7 style app receipts that contain auto-renewable or non-renewing subscriptions.* If value is true, response includes only the latest renewal transaction for any subscriptions. */
+  'exclude-old-transactions'?: boolean
+}
+
+export interface ReceiptValidationResponse {
```
cometkim: Missing `is-retryable` property. 

> `is-retryable`: Retry validation for this receipt. Only applicable to status codes 21100-21199 (listed in Table 2-1)

index.d.ts
cometkim: That's good plan. But breaks compatibility, so I think we'll have to deal with this change separately.

LinusU: > But breaks compatibility

This change wouldn't though? Because there are no `SkuTypes` actually exported from `index.js`?

I still think that we should change it though, shouldn't be a problem to introduce breaking changes since we are in an alpha release?

cometkim: Ah, sorry. I confused. I meant the plan. This change looks good.

But other changes below:
```diff
export interface Product {
-  type: SkuTypeAndroid | SkuTypeIOS;
+ type: 'inapp' | 'iap'
```
This is the plan you said right?

LinusU: > It looks typeof getProducts is not valid for Android until the SkuType is nomalized.

I believe it is, `'inapp'` is for Android and `'iap'` for iOS.

Note that I split up `Product` and `Subscription` since `getProducts` only returns products (I think at some point in history it also included subscriptions on iOS 🤔)

cometkim: I'm not sure about. Alright, If it's case insensitive. :ok_hand:

dooboolab:
@cometkim @LinusU  Thank for your work!

cometkim:
Hey @dooboolab and @LinusU, I was noticed about "Mapped types on tuples" support in TypeScript 3.1

It looks like we can make this declaration be exact and short, but should drop support below versions of TS.

```ts
export function getProducts<A extends string, B extends string, C extends string, D extends string, E extends string, F extends string>(skus: [A, B, C, D, E, F]): Promise<[Product<A>, Product<B>, Product<C>, Product<D>, Product<E>, Product<F>]>;
export function getProducts<A extends string, B extends string, C extends string, D extends string, E extends string>(skus: [A, B, C, D, E]): Promise<[Product<A>, Product<B>, Product<C>, Product<D>, Product<E>]>;
export function getProducts<A extends string, B extends string, C extends string, D extends string>(skus: [A, B, C, D]): Promise<[Product<A>, Product<B>, Product<C>, Product<D>]>;
export function getProducts<A extends string, B extends string, C extends string>(skus: [A, B, C]): Promise<[Product<A>, Product<B>, Product<C>]>;
export function getProducts<A extends string, B extends string>(skus: [A, B]): Promise<[Product<A>, Product<B>]>;
export function getProducts<A extends string>(skus: [A]): Promise<[Product<A>]>;
export function getProducts(skus: string[]): Promise<Product<string>[]>;
```

How do you think about?

LinusU:
> [...] but should drop support below versions of TS.

works for me 👍

dooboolab:
@cometkim @LinusU Actually, I don't know how to manage this one since it isn't downward compatible. Would this be ok to just drop the lower version support?

cometkim:
@dooboolab It's not that important while it is not a dependency or peer dependency.

cometkim:
@LinusU I just tried to use tuple mapping, but `keyof` does not support explicit inference of arrays or tuple entries.. it looks always `string | number | symbol` and the `number` is incompatible with `Product<ID extends string>`.

Closed this topic..

index.d.ts
```diff
@@ -65,17 +63,27 @@ export function consumeAllItems() : Promise<void>;
...
+ * @param skus The item skus
  */
-export function getProducts(skus: string[]) : Promise<Product[]>;
+export function getProducts<A extends string, B extends string, C extends string, D extends string, E extends string, F extends string>(skus: [A, B, C, D, E, F]): Promise<[Product<A>, Product<B>, Product<C>, Product<D>, Product<E>, Product<F>]>;
+export function getProducts<A extends string, B extends string, C extends string, D extends string, E extends string>(skus: [A, B, C, D, E]): Promise<[Product<A>, Product<B>, Product<C>, Product<D>, Product<E>]>;
+export function getProducts<A extends string, B extends string, C extends string, D extends string>(skus: [A, B, C, D]): Promise<[Product<A>, Product<B>, Product<C>, Product<D>]>;
+export function getProducts<A extends string, B extends string, C extends string>(skus: [A, B, C]): Promise<[Product<A>, Product<B>, Product<C>]>;
+export function getProducts<A extends string, B extends string>(skus: [A, B]): Promise<[Product<A>, Product<B>]>;
+export function getProducts<A extends string>(skus: [A]): Promise<[Product<A>]>;
+export function getProducts(skus: string[]): Promise<Product<string>[]>;
```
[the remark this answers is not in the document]
LinusU: It makes `getProducts(['a', 'b'])` return a tuple of `Product<'a'>` and `Product<'b'>`, instead of an array of generic products.

That way the following will be correctly caught as errors:

```ts
const products = await getProducts(['a', 'b'])

products[2] // TS will complain since it's only a 2 element tuple

if (products[0].productId === 'c') // TS will complain since none of the productIds can be 'c'
```

PR #240, opened 2018-08-16
Add buyWithQuantity iOS function

lefoy:
This PR add the buyWithQuantity iOS only function. I also added an optional parameter to the buyProductWithoutFinishTransaction since it's already iOS only.


index.js
```diff
@@ -104,15 +104,27 @@ export const buyProduct = (sku) => Platform.select({
...
+/**
+ * Buy a product with a specified quantity (iOS only)
+ *   Call finishTransaction after receipt validation process.
+ * @param {string} sku The product's sku/ID
+ * @param {number} quantity The amount of product to buy
+ * @returns {Promise<ProductPurchase>}
+ */
+export const buyProductWithQuantity = (sku, quantity) => Platform.select({
+  ios: () => RNIapIos.buyProductWithQuantity(sku, quantity),
+  android: () => RNIapModule.buyItemByType(ANDROID_ITEM_TYPE_IAP, sku, null)
```
dooboolab: I think it would be better to just resolve in `android`.
```
() => Promise.resolve(),
```

Just make above method only `ios` dependent. Naming it as `buyProductWithQuantityIOS` will be better too.

dooboolab:
@lefoy Could you review above and give some changes in naming?

index.js
```diff
@@ -104,15 +104,27 @@ export const buyProduct = (sku) => Platform.select({
...
 /**
  * Buy a product without transaction finish (iOS only)
  *   Call finishTransaction after receipt validation process.
  * @param {string} sku The product's sku/ID
+ * @param {number} quantity The amount of product to buy (optional, default = 1)
  * @returns {Promise<ProductPurchase>}
  */
-export const buyProductWithoutFinishTransaction = (sku) => Platform.select({
-  ios: () => RNIapIos.buyProductWithoutAutoConfirm(sku),
+export const buyProductWithoutFinishTransaction = (sku, quantity = 1) => Platform.select({
```
dooboolab: Also, I wish not to control the codes that affect both platforms. Hope this to be rollbacked and if you really want this, I think it'd be better to make different method.

index.js
lefoy: I agree with `buyProductWithQuantityIOS`, I will update my PR

index.js
lefoy: I'm not sure I understand your comment, `buyProductWithoutFinishTransaction` is already documented as iOS only

index.js
lefoy: Actually, if the method `buyProductWithoutFinishTransaction` is already iOS only, we should either rename both method or leave `buyProductWithQuantity` what do you think?

dooboolab: @lefoy Good point. That might have confused you. `buyProductWithoutFinishTransaction` however works seamlessly with android. The purpose of this method is to avoid omission in response since there can be delays in callback from apple itself waiting for the transaction to complete. Using this feature may be risky in some situation, but some devs wanted it for whatever reason. 
Anyway, I want to confirm you that `buyProductWithoutFinishTransaction` share almost the same feature in each platform with some differences at the back-side. Therefore, it is somewhat different from `buyProductWithQuantity` feature because it actually is serving different parameter that won't be used in `android`.
Therefore, I think it'd be better to name it as `buyProductWithQuantityIOS` and not provide it for android.

index.js
dooboolab: @lefoy Yes, that may confuse you as I described above. However, that actually works with `android` however. Therefore, I thought it'd be better not to modify codes in `buyProductWithoutFinishTransaction`. I just wanted your feature to be completely dependent to prevent code and features tangling as this module keeps digesting new features.

lefoy:
@dooboolab The PR should be ready to merge now, let me know if I missed something

dooboolab:
@lefoy Looks good. I'll merge this and publish to `2.0.0-alpha8`.

lefoy:
Thank you! 👍

Issue #236, opened 2018-08-15
Wishlist for 2.0


LinusU:
Just started using this library and noticed a few shortcomings that I would love to get fixed before 2.0 is out as stable:

- [x] Remove `RNVersion` as parameter from `validateReceipt*` (just read this from `react-native` or simply drop support for old versions)
- [ ] Proper error handling
  - [ ] Add `code` property to all errors
  - [x] Let `prepare` throw an error when payments are disabled on iOS

    Currently, this method is documented as 1) being a noop and 2) returning `Promise<void>`. The reality is that it returns the string "false" on iOS when payments aren't enabled.

    I think it should return `Promise<void>` on iOS, but reject with an error (e.g. `PAYMENTS_NOT_ENABLED`) when payments aren't enabled.
  - ~~Let `validateReceiptIos` reject on Android (e.g. `WRONG_PLATFORM`)~~ (better to have it just work on both platforms)
  - [x] Let `validateReceiptIos` reject instead of returning false and printing to console
- [x] Fix `validateReceiptAndroid` (it always returns `false` for React Native >= 0.54)
- [ ] Update documentation for `getProduct` and `getSubscription` to remove note about not making a difference with product/subscriptions (this have been fixed with a `.filter()` in the code)
- [x] Possibly rename `prepare` and `endConnection` to better match each other
  - e.g. `initConnection`/`endConnection` or `prepare`/`teardown` or something else entirely
- [ ] Abstract away the different types on iOS/Android (e.g. `inapp` vs `iap`), just make sure that `type` is the same on both platforms.
- [ ] Abstract away the differences between `price`/`localizedPrice` on iOS/Android
  - Currently `price` is localized on Android, but not on iOS
  - Would like to change so that `price` is a `number`, and `localizedPrice` is `string`

Most of these changes are breaking changes which is why I would love to get them in before final 2.0.

I would be happy to submit pull requests for all of the items above 🌻 😎

dooboolab:
@LinusU Thanks. We will make this as a todo list before submitting to `2.0`. Will love your `PR` too.

LinusU:
Nice! Thanks!

Feel free to start with #235 ☺️

LinusU:
#238 addresses the first point

LinusU:
(added point about `price`/`localizedPrice` to first post)

LinusU:
@dooboolab do you want to merge #235 and #238, and tag a new alpha? Would love to get those changes in ☺️ 

[...]

PR #242, opened 2018-08-18
Improve receipt validation

LinusU:
Got quite a lot of things in this PR, easiest is to go commit by commit when reviewing:

- Add back return statement that got lost in merge
- Improve validateReceiptAndroid
- Added a package lock file (npm kept complaining 😆)
- Add eslint and fix all linting errors (uses config already present in repo, re-enabled some disabled rules that we did not break)
- Updated the readme around getSubscriptions on iOS

Realated to #236


dooboolab:
@LinusU Wow clean code! Will merge this now.

Issue #261, opened 2018-09-13
Android: onHostDestroy checks if mServiceConn is null when it should check if mService is null


fierysolid:
### Version of react-native-iap
react-native-iap: 2.1.2
react-native: 0.57.0

### Platforms you faced the error (IOS or Android or both?)
Android

### Expected behavior
The app shouldn't immediately crash
### Actual behavior
The app crashes immediately
### Tested environment (Emulator? Real Device?)
Galaxy S7 Edge (Real Device)
### Steps to reproduce the behavior
My app only calls `.initConnection()` when a user opens my IAP Modal, then immediately closes the connection again when it closes. For some reason, the phone immediately calls `onHostDestroy` which checks if `mServiceConn` is null before trying to unbind the service. `mServiceConn` will always not be null because its an object. Instead, we need to check if `mService` is null as that is the indicator of whether the service has been connected or not and whether or not we should unbind.

This line:
https://github.com/dooboolab/react-native-iap/blob/master/android/src/main/java/com/dooboolab/RNIap/RNIapModule.java#L93

should be changed to `mService != null`

dooboolab:
Merged! Released in `2.1.3`.

PR #262, opened 2018-09-13
onHostDestroy fix

fierysolid:
Checking if `mService` is null instead of `mServiceConn` before trying to unbind service. `mServiceConn` will always be an object.


fierysolid:
This is a PR for [261](https://github.com/dooboolab/react-native-iap/issues/261)

Issue #257, opened 2018-09-12
iOS: This in-app purchase has already been bought.


siderakis:
### Version of react-native-iap
    "react-native-iap": "^2.0.3",

### Platforms you faced the error (IOS or Android or both?)
iOS

### Expected behavior
Makes a purchase

  const purchase = await RNIap.buyProductWithoutFinishTransaction(productId)
      const { transactionReceipt, purchaseToken } = purchase
      sendToServer(transactionReceipt,        {
          onSuccess: () => {
            RNIap.finishTransaction()
            RNIap.consumeAllItems()
          }})

### Actual behavior
Dialog: “This in-app purchase has already been bought”. This is only happening for one purchase, others are working fine. I think the app got in a bad state and I don't see a way to fix it. Other SKUs I can purchase multiple times without issue.

### Tested environment (Emulator? Real Device?)
Device (sandbox)

dooboolab:
@siderakis Yeah strange. I've already patched this problem like told in [stackoverflow](https://stackoverflow.com/questions/34001868/ios-this-in-app-purchase-has-already-been-bought-pop-up) but still occurring in some devices. I am willing to fix this too. What device are you using?

dooboolab:
@siderakis [This](https://stackoverflow.com/questions/18210747/how-to-remove-a-very-old-defunct-in-app-purchase-from-the-transaction-queue) might solve your app in bad state I guess. Could you try?

siderakis:
I tried it and it didn't work. I also reinstalled the app and that didn't help. It's running on an iPad.

siderakis:
I set a breakpoint and it looks like currentQueue.transactions is empty

siderakis:
From SKPaymentQueue.h

```
// Array of unfinished SKPaymentTransactions.  Only valid while the queue has observers.  Updated asynchronously.
@property(nonatomic, readonly) NSArray<SKPaymentTransaction *> *transactions NS_AVAILABLE_IOS(3_0);

@end

```

Maybe I called it too soon

siderakis:
I called the code later in the programs execution and it fixed the issue.

JJMoon:
@siderakis Sorry for bothering. But, what code did you call ?

JJMoon:
@siderakis Yes. That's what I meant in issue #259 .
I think that action can be anywhere after applicationDidBecomeActive.
So, I will expose that method to JS side, that user can call it any time.
Thank you for your advice.

[...]

PR #264, opened 2018-09-16
Force finish transaction

JJMoon:
Calling finishTransaction, which I mentioned, would be no effect when restarting the App or device itself. So, I looked up the document on iOS I found `transactions` property in skpaymentqueue class, and I used it for this implementation.

https://developer.apple.com/documentation/storekit/skpaymentqueue

I named the new method to `clear transaction` which looks fit better.
Check and test the method. Especially the jsDoc return values etc.

### DIFF
--- a/index.js
+++ b/index.js
@@ -149,6 +149,17 @@ export const finishTransaction = () => Platform.select({
 })();
 
 /**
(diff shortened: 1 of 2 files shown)


dooboolab:
resolved #257

PR #267, opened 2018-09-18
Automatically handle connection initialization

LinusU:
The diff might be a bit hard to read but it's mostly indentation changes 🙈 

Basically, I've wrapped all calls with a new `ensureConnection` function. This function will detect if the client hasn't been initialized, and in that case initialize it before running the supplied code.

The end users can continue to call `initConnection` and `endConnection` as before, but it's now optional since a connection will be automatically established if it's needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,8 @@ Also, note that this is our last migration for renaming method names without any
 #### Methods
 | Func  | Param  | Return | Description |
 | :------------ |:---------------:| :---------------:| :-----|
-| prepare |  | `Promise<boolean>` | Deprecated. Use `initConnection instead` |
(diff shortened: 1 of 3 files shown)


dooboolab:
@LinusU Do you think I can just publish this to new version?

LinusU:
Should be good to go, currently using it in an app live on the play store: https://play.google.com/store/apps/details?id=com.ctrlpanel

The changes should be backwards compatible since you can still intialize and close the connection manually.

dooboolab:
@LinusU What an awesome app! I'll publish this to 2.2.2. Thank you!

LinusU:
Thank you! ❤️

PR #270, opened 2018-09-25
[WIP] Android Subscription Proration

fierysolid:
Work in Progress. Found a bug with downgrades


dooboolab:
@fierysolid Is this ready?

fierysolid:
Not yet, still working out some kinks. Should have it ready EOD though.

dooboolab:
@fierysolid NP you may slow it down. Tell me when it is ready, I'll review this.

android/src/main/java/com/dooboolab/RNIap/RNIapModule.java
```diff
@@ -347,7 +347,7 @@ public void onPurchaseHistoryResponse(@BillingClient.BillingResponse int respons
   }
 
   @ReactMethod
-  public void buyItemByType(final String type, final String sku, final String oldSku, final Promise promise) {
+  public void buyItemByType(final String type, final String sku, final String oldSku, final Integer prorationMode, final Promise promise) {
     final Activity activity = getCurrentActivity();
```
dooboolab: @fierysolid You need to be aware that `public void buyItemByType(final String type, final String sku, final String oldSku, final Promise promise)` and `public void buyItemByType(final String type, final String sku, final String oldSku, final Integer prorationMode, final Promise promise)` are different method in java. You should make another method and leave `buyItemByType` as it was before. Java isn't working like javascript and it actually treats method with the different number of parameters differently.

PR #280, opened 2018-10-06
Provide dummy value for required argument 'proprationMode' when invok…

mars-lan:
…ing RNIapModule.buyItemByType

Failing to due so results in immediate crash when calling buyProudct


dooboolab:
Thank you for the fix and sorry for the breakings in `2.3.0`.

PR #282, opened 2018-10-07
Replace deprecated 'compile' gradle configuration with 'implementation'

mars-lan:
See http://d.android.com/r/tools/update-dependency-configurations.html for more details

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -46,6 +46,6 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


mars-lan:
I believe this is already the case for RN 0.57? 

https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#057

dooboolab:
@mars-lan Oh. I've missed that thank you. I will merge this!

PR #285, opened 2018-10-12
Add missing field in ProductPurchase interface

raynor85:
`RNIap.buyProduct` also returns `purchaseToken` on Android devices, but it is missing in `ProductPurchase` interface


dooboolab:
Thank you!

Issue #290, opened 2018-10-20
Why use custom error messages?


mars-lan:
What's the rationale for manually mapping error codes to custom descriptions here: 
https://github.com/dooboolab/react-native-iap/blob/98e9d230be99ce87f368e0446ed108656b0588b8/ios/RNIapIos.m#L311

Wouldn't it better to simply use whatever returned by https://developer.apple.com/documentation/foundation/nserror/1414418-localizeddescription?language=objc  ? This also has the benefit of localizing the error messages.

dooboolab:
Yeah maybe you can give a `PR` for this one.

PR #291, opened 2018-10-20
Use NSError's localizedDescription instead of custom error messages

mars-lan:
Fixes https://github.com/dooboolab/react-native-iap/issues/290


dooboolab:
Thanks for the fix. @lwansbrough if you are available, could you review shortly, if this is ok?

lwansbrough:
If I recall correctly, the built in error descriptions are bad.

dooboolab:
@lwansbrough Thank you for the review. What do you think @mars-lan ?

mars-lan:
While the built-in error message may not be as informative, at least it's localized. Also trying to maintain a positional-based mapping like this seems very fragile to me. 

Since the error code is also returned, if really needed, clients can use that to display custom error messages on JS side.

dooboolab:
@mars-lan Alright. I agree with you. Let's merge this and see how it goes.

Issue #279, opened 2018-10-05
buySubscription ignoring user feedback


marcosmartinez7:
### Version of react-native-iap

2.3.0
### Platforms you faced the error (IOS or Android or both?)
Android
### Expected behavior

let sub = await RNIApp.buySubscription("some");

should wait for the user to press buy on the modal of in app payments

### Actual behavior

i cant check if the user accepted the subscription, because when doing:

let sub = await RNIApp.buySubscription("some");
if(sub){
   do some things..
}

the if(sub) block executes inmediatly after the popup opens, but the user still not accept the subscription

### Tested environment (Emulator? Real Device?)

Real device

Any ideas?

There  are other methods like getAvailablePurchases that also returns a promise and works fine

The problem is that buySubscription should return the callback of the user input and it isnt

marcosmartinez7:
Literally didnt touch anything, and after some uninstall/install it works, any ideas if this must be related with the playstore working wrong? 

Its the only logical explanation that i can imagine ..

dooboolab:
@marcosmartinez7 Sorry for the breaking could you try `2.3.1`?

marcosmartinez7:
Hi, thanks! 

I dont have time right now but i will

dooboolab:
@marcosmartinez7 Thanks please come back.

Gaia-Nutrition:
@dooboolab Issue is present on iOS too. Tried both `2.3.1` as well as `2.3.2`

dooboolab:
I think it is fixed in `2.3.3`. Please reopen when this is still occurring.

Gaia-Nutrition:
@dooboolab I'm sorry, just installed and relinked `2.3.5`. `await RNIap.buySubscription(iapProductId)` still immediately returns `undefined` on iOS

Gaia-Nutrition:
@dooboolab I tried linking and unlinking. Whenever I switch back to `2.2.2` it starts working again. But I'll check again tomorrow. Maybe try to manually install the native component

Gaia-Nutrition:
@dooboolab Ok, I now manually linked the new `2.3.6` version. Still getting `undefined`. I'd love to help solve this issue. Let me know what I can do to help!

dooboolab:
Can somebody confirm this? This shouldn't happen.

freakyfriday:
[...]

PR #295, opened 2018-10-23
Fix for missing async await.

JJMoon:
Add missing async, await notation. 
이거 안드로이드는 괜찮은 지 함 보셔요..

### DIFF
--- a/index.js
+++ b/index.js
@@ -104,7 +104,7 @@ export const getAvailablePurchases = () => Platform.select({
  */
 export const buySubscription = (sku, oldSku, prorationMode) => {
   return Platform.select({
-    ios: () => RNIapIos.buyProduct(sku),
+    ios: async () => await RNIapIos.buyProduct(sku),
     android: () => {
       if (!prorationMode) prorationMode = -1;
       return RNIapModule.buyItemByType(ANDROID_ITEM_TYPE_SUBSCRIPTION, sku, oldSku, prorationMode);
@@ -118,7 +118,7 @@ export const buySubscription = (sku, oldSku, prorationMode) => {
  * @returns {Promise<ProductPurchase>}
  */
 export const buyProduct = (sku) => Platform.select({
-  ios: () => RNIapIos.buyProduct(sku),
+  ios: async () => await RNIapIos.buyProduct(sku),
   android: () => RNIapModule.buyItemByType(ANDROID_ITEM_TYPE_IAP, sku, null, 0),
 })();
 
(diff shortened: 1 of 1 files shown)


dooboolab:
This should be done to android too.
[LINK](https://facebook.github.io/react-native/docs/native-modules-android)

dooboolab:
Resolve #279

dooboolab:
@JJMoon Can you read [this](https://stackoverflow.com/questions/44806135/why-no-return-await-vs-const-x-await)? I don't quite get why this process resolve #279.

dooboolab:
@lwansbrough Hi. I need your help on this `PR`. Do you think we need to call another `async` and `await` to ensure the finishes of the `function`? I feel like it is quite redundant. #279 was resolved with doing this fixes but I am still not sure why we need this. Or could be a better way.

LinusU:
This removes the return value from these functions, there was nothing wrong with them before.

Could this please be reverted?

hyochan:
@LinusU I felt this is quite weird too. However, @JJMoon swear to god that this should be done to fix #279. I've just looked into this one and feels like this wasn't the solution and it was just `fragile linking` problem which we should just `unlink` and `link` again. What do you think @LinusU @JJMoon ?

LinusU:
It's possible that Babel is rewriting the async function in some way that makes the Promise resolve later in the microtask scheduling queue than before, but I don't see how that could possibly affect anything 🤔

hyochan:
@LinusU Ok. I'll try to revert this tomorrow if I don't receive any other feedback on this. Especially from @JJMoon.

hyochan:
@LinusU Wouldn't this be fine tough?
```
async () => RNIapIos.buyProduct(sku),
```
I've just looked in the code and saw that I've already gave some updates.

LinusU:
The `async` there won't do anything, I would recommend keeping it as `() => RNIapIos.buyProduct(sku)`

hyochan:
@LinusU Yes. This is just another `Promise resolver`. Ok I'll change this now then.

hyochan:
@LinusU Done it!. @JJMoon Please check if you have any opinion on this release which you've told me to make changes. Thanks.

JJMoon:
@hyochan It looks my PR is merged. It's OK. Thanks

hyochan:
@JJMoon That wasn't what I was saying. The changes you've made which you swear to god to make all the functions to `async`, we thought this should be reverted. Please read the previous conversations. Thanks.

PR #311, opened 2018-11-06
resolve(null) => resolve(true)

eminsr:
In this project, if a promise is not rejected it is successful so returning null was a bit confusing on the Android side. Now it is more consistent with ios.

`const result = await RNIap.initConnection();`
Now the result is `true`


dooboolab:
Yeah. I agree. Merged!

PR #344, opened 2018-12-20
Expose BCP47 language tag for iOS.

m-sanders:
Expose BCP47 language tag for iOS. This value can then be used in conjunction with ECMAScript I18n API `Intl.NumberFormat` to format currency correctly. #342 

See:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat
https://developer.apple.com/documentation/foundation/nslocale/1416263-localeidentifier
https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html

PR #345, opened 2018-12-20
Revert "Expose BCP47 language tag for iOS."

hyochan:
Reverts dooboolab/react-native-iap#344


DataGreed:
why was it reverted?

DataGreed:
@hyochan why was it reverted? It's a crucial feature

mikefogg:
@hyochan Just wanted to check on this, is there a reason it was reverted? I need to get annual pricing in terms of months (so annual / 12) and need access to the locale (or some other method for calculating price like that). Wanted to make sure there wasn't a specific reason before I try and add it back in :)

hyochan:
I have reverted this since it broke the previous build. Any further update can be given as long as it works on example project.

Issue #315, opened 2018-11-09
Android crashes in many devices after release


umutpiri:
### Version of react-native-iap
 "react-native": "0.55.4"
"react-native-iap": "^2.3.2"
### Platforms you faced the error (IOS or Android or both?)
Android
### Expected behavior
Please help me, I need to fix this immediately. Because there are already +2k users
### Actual behavior
App crashes because of rniap in many devices. In play console, I saw this error causes the crash : 
java.lang.RuntimeException: 
1.  at com.facebook.react.bridge.CallbackImpl.invoke (CallbackImpl.java:28)
2. at com.facebook.react.bridge.PromiseImpl.resolve (PromiseImpl.java:30)
3.  at com.dooboolab.RNIap.RNIapModule$4.run (RNIapModule.java:154)
4.  at com.dooboolab.RNIap.RNIapModule$3.onBillingSetupFinished (RNIapModule.java:123)
5.  at com.android.billingclient.api.BillingClientImpl$BillingServiceConnection.onServiceConnected (BillingClientImpl.java:903)
6.  at android.app.LoadedApk$ServiceDispatcher.doConnected (LoadedApk.java:1264)
7.  at android.app.LoadedApk$ServiceDispatcher$RunConnection.run (LoadedApk.java:1281)
8.  at android.os.Handler.handleCallback (Handler.java:815)
9.  at android.os.Handler.dispatchMessage (Handler.java:104)
10.  at android.os.Looper.loop (Looper.java:207)
11.  at android.app.ActivityThread.main (ActivityThread.java:5692)
12.  at java.lang.reflect.Method.invoke (Native Method)
13.  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:908)
14.  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:769) 
### Tested environment (Emulator? Real Device?)
here is the report from play console : 
https://ibb.co/gDxZQA  
### Steps to reproduce the behavior
I don't know what to do, I used rniap like in example. 
in componentDidMount : 
try {
      const result = await RNIap.initConnection();
    } catch (err) {
      console.log(err);
    }
in componentWillUnmount: 
RNIap.endConnection();

umutpiri:
I can't. Because it was an immediate situation. So, I changed the iap package. Now there are no crashes in +3k users. But, thank you. I will try it in my next project.

Ilario17:
[...]

PR #359, opened 2019-01-07
Do not null mBillingClient (#315)

mitchellmcm27:
This PR tries to solve issues brought up in #315 , specifically that methods are invoked on `mBillingClient` when it is `null`. See my comment there for a discussion of the problem. Copy-pasted here, the variable `clientReady` is being used to track if the billing client setup has been completed, by setting it to `false` in the onBillingServiceDisconnected() callback. But `mBillingClient` itself is set to `null` immediately when `endConnection()` is called. This is a mismatch. It's possible for `clientReady` to remain set to `true`, even if `mBillingClient` is `null`. In that case, methods will be invoked by the on the null object instance. It seems that something is preventing `onBillingServiceDisconnected()` from being called.

Solution:

- avoid setting `mBillingClient` to `null`. It is still not initialized in the class constructor, so will be `null` before `ensureConnection()` is called for the first time. Thus the null checks are still necessary.

- Remove the `clientReady` property in favor of relying on `mBillingClient.isReady()` (again checking for null before calling).

The root cause of this issue is that when `endConnection()` is called as a React Native component is unmounting, it seems that the `onBillingServiceDisconnected()` callback is never invoked. I don't know whether this is intended by the API or not, so I did not attempt to change this. I suggest someone with more experience in Android In App Billing review this PR before it is merged.

### DIFF
--- a/android/src/main/java/com/dooboolab/RNIap/RNIapModule.java
+++ b/android/src/main/java/com/dooboolab/RNIap/RNIapModule.java
@@ -1,3 +1,4 @@
+
 package com.dooboolab.RNIap;
 
 import android.app.Activity;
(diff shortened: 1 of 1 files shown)


hyochan:
Thanks for the `PR` and the description. Looks like it has been rollbacked. However, since I agree that this isn't the solution. I will merge this now. Thank you.

PR #363, opened 2019-01-12
fix example to correctly display available purchases

jrhee17:
[GIVEN]
- Location: Iap example app
- Issue: Button labelled `Get available purchases` gets items instead, which is confusing for newbies like myself:(

[THEN]
- Fix button to correctly display available purchases

Not sure what the correct flow for PRs are -- let me know if there are any issues

### DIFF
--- a/RNIapExample/src/components/pages/First.js
+++ b/RNIapExample/src/components/pages/First.js
@@ -141,7 +141,7 @@ class Page extends Component {
           >
             <View style={{ height: 50 }} />
             <NativeButton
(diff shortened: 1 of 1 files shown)


hyochan:
@JJMoon Please review this.

JJMoon:
@jrhee17 Thank you for work. But, I wonder we still need another button for `getItems`.
Just in iOS only. I guess I need to test both methods.

hyochan:
@jrhee17 I've just see this and your are right.

PR #378, opened 2019-01-23
Fix: add return type to TS definition of 'addAdditionalSuccessPurchaseListenerIOS'

### LINKED ISSUE #376: [TypeScript] `addAdditionalSuccessPurchaseListenerIOS` lacking return type definition

2.4.0-beta6

### Version of react-native

Any

Both

### Details

`addAdditionalSuccessPurchaseListenerIOS`'s definition is lacking a return type in [index.d.ts](https://github.com/dooboolab/react-native-iap/blob/master/index.d.ts#L193) which renders the following error when `noImplicitAny` is enabled on TypeScript:

```
node_modules/react-native-iap/index.d.ts(193,17): error TS7010: 'addAdditionalSuccessPurchaseListenerIOS', which lacks return-type annotation, implicitly has an 'any' return type.
```

### Solution

According to [index.js#242](https://github.com/dooboolab/react-native-iap/blob/4eb8ab9ac4867e291501d7649902ec4be8cc9080/index.js#L242), `addAdditionalSuccessPurchaseListenerIOS` returns a EmitterSubscription, since it passes forward the return value of a call to `NativeEventEmitter.addListener`.

Therefore, the type definition should be:

``` ts
export function addAdditionalSuccessPurchaseListenerIOS(fn: Function): EmitterSubscription;
```

fjcaetano:
Declares the function's return type as `EmitterSubscription`

fix #376

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,5 @@
+import { EmitterSubscription } from 'react-native';
+
 import * as Apple from './apple'
 
 interface Common {
@@ -190,4 +192,4 @@ export function validateReceiptAndroid(packageName: string, productId: string, p
   * Add IAP purchase event in ios.
  * @returns {callback(e: Event)}
  */
-export function addAdditionalSuccessPurchaseListenerIOS(fn: Function);
+export function addAdditionalSuccessPurchaseListenerIOS(fn: Function) : EmitterSubscription;


hyochan:
Thank you!

PR #388, opened 2019-01-29
Add support for iOS In-App Purchase promotions

### LINKED ISSUE #199: Promote IAPs from App Store?
Does this plugin support SkPaymentTransactionObserver method so that the In-App Purchases can be promoted on the iOS App Store?
https://developer.apple.com/app-store/promoting-in-app-purchases/

neilco:
This pull request adds support for the additional `SKProductTransactionObserver` delegates that support promoting In-App Purchase products in the App Store. 

This pull request closes #199.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,7 +8,7 @@ const ANDROID_ITEM_TYPE_IAP = 'inapp';
 const IOS_ITEM_TYPE_SUBSCRIPTION = 'sub';
 const IOS_ITEM_TYPE_IAP = 'iap';
 
-export const onPromotedProduct = 'iap-promoted-product';
+export const PROMOTED_PRODUCT = 'iap-promoted-product';
 
 /**
  * @deprecated Deprecated since 2.0.0. Use initConnection instead.
@@ -194,8 +194,8 @@ export const consumePurchase = (token) => Platform.select({
  */
 export const getPromotedProduct = () => Platform.select({
   ios: async() => RNIapIos.promotedProduct(),
-  android: async() => Promise.resolve()
-})()
+  android: async() => Promise.resolve(),
+})();
 
 /**
  * Buy the currently selected promoted product (iOS only)
@@ -204,8 +204,8 @@ export const getPromotedProduct = () => Platform.select({
  */
 export const buyPromotedProduct = () => Platform.select({
   ios: async() => RNIapIos.buyPromotedProduct(),
-  android: async() => Promise.resolve()
-})()
+  android: async() => Promise.resolve(),
+})();
 
 /**
  * Validate receipt for iOS.
--- a/index.js
+++ b/index.js
@@ -8,6 +8,8 @@ const ANDROID_ITEM_TYPE_IAP = 'inapp';
 const IOS_ITEM_TYPE_SUBSCRIPTION = 'sub';
 const IOS_ITEM_TYPE_IAP = 'iap';
 
+export const onPromotedProduct = 'iap-promoted-product';
+
 /**
  * @deprecated Deprecated since 2.0.0. Use initConnection instead.
  * @returns {Promise<void>}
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -8,6 +8,8 @@ const ANDROID_ITEM_TYPE_IAP = 'inapp';
 const IOS_ITEM_TYPE_SUBSCRIPTION = 'sub';
 const IOS_ITEM_TYPE_IAP = 'iap';
 
+export const onPromotedProduct = 'iap-promoted-product';
```
hyochan: I think you can delete this one?

index.js
```diff
@@ -185,6 +187,26 @@ export const consumePurchase = (token) => Platform.select({
...
 
+/**
+ * Should Add Store Payment (iOS only)
+ *   Indicates the the App Store purchase should continue from the app instead of the App Store. 
+ * @returns {null}
+ */
+export const getPromotedProduct = () => Platform.select({
+  ios: async() => RNIapIos.promotedProduct(),
+  android: async() => Promise.resolve()
+})()
```
hyochan: Hope you but `;` <== at the end to follow other lint.

index.js
```diff
@@ -185,6 +187,26 @@ export const consumePurchase = (token) => Platform.select({
...
+
+/**
+ * Buy the currently selected promoted product (iOS only)
+ *   Initiates the payment process for a promoted product. Should only be called in response to the `iap-promoted-product` event.  
+ * @returns {null}
+ */
+export const buyPromotedProduct = () => Platform.select({
+  ios: async() => RNIapIos.buyPromotedProduct(),
+  android: async() => Promise.resolve()
+})()
```
hyochan: same `lint` fix like above.

index.js
hyochan: Also, we are using `trailing commas`.

neilco: It might be worth setting up ESLint and adding your style rules to the repo rather than going through this process each time ;)

ios/RNIapIos.m
```diff
@@ -35,6 +38,14 @@ +(BOOL)requiresMainQueueSetup {
   return YES;
 }
 
+- (void)startObserving {
```
hyochan: Where is this method used? Also `stopObserving`?

index.js
```diff
@@ -185,6 +187,26 @@ export const consumePurchase = (token) => Platform.select({
   android: async() => RNIapModule.consumeProduct(token),
 })();
 
+/**
+ * Should Add Store Payment (iOS only)
+ *   Indicates the the App Store purchase should continue from the app instead of the App Store. 
+ * @returns {null}
+ */
+export const getPromotedProduct = () => Platform.select({
```
hyochan: If you are supporting only `iOS` currently, please rename this `getPromotedProductIOS` in order to make users clear that only `iOS` is supported.

hyochan:
@neilco Thank you for the contribution. This is really great support!. I've reviewed some of your lines.

index.js
neilco: One could argue that since it’s a NOP on Android, so it’s not only iOS that’s supported. Personally I would prefer NOT to use platform suffixes. There’s also existing precedent that a suffix is not required—see `clearProducts`, `clearTransaction`, `consumePurchase`, `finishTransaction`, etc.

index.js
neilco: Sure, but a well-defined constant is far better than a bare string at the point of use.

hyochan: Then I think the const variable should be renamed to something like `PROMOTED_PRODUCT'.

index.js
hyochan: There is actually defined [here](https://github.com/dooboolab/react-native-iap/blob/master/.eslintrc) however this is pretty outdated and should be updated. However, [comma-dangle](https://github.com/dooboolab/react-native-iap/blob/3914f6f3a3f3a7d69d74f78b2709de70752793ea/.eslintrc#L8) defined here.

ios/RNIapIos.m
hyochan: Wow. Thanks. This is new to me.

index.js
hyochan: Yeah. You are right. This is quite not well organized yet. Leaving this as `NOP` would be fine. Thank you for your reasoning.

neilco:
@hyochan All ready for review 🙈

kristfal:
Question for this: How would one go about getting the full receipt for a promoted product after the purchase? This is relevant for receipt validation.

neilco:
@kristfal Purchased promoted products are the same as regular purchases so handle the receipt in exactly the same way.

kristfal:
Ok, thanks for info. Reason I was asking is because the return value of `buyPromotedProduct` is not documented. Is it of type `<Purchase>` (same as `buyProduct`/`buySubscription`?

neilco:
@kristfal `buyPromotedProduct` returns a Promise that is resolved with [a subset of the purchase data, including the transaction receipt](https://github.com/dooboolab/react-native-iap/blob/master/ios/RNIapIos.m#L482). This is exactly the same as `buyProduct` and `buySubscription`. 

The only difference is that the product choice happens external to the app—the `IAPPromotedProductObserver` receives a `SKProduct` and `SKPayment` from the App Store and `buyPromotedProduct` adds that `SKPayment` to the `SKPaymentQueue`.

PR #431, opened 2019-03-19
Start adding Flow type definitions

pietro909:
Hello,
thank you for the great package!

We've been using custom Flow definitions for the last ~3 months and I think they are now stable and reliable enough to be added to the codebase.

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -1,0 +1,97 @@
+// @flow
+
+declare type Common = {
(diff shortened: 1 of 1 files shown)


hyochan:
Thanks. We now have both `ts` and `flow`.

hyochan:
Released to `2.4.3`.

Issue #409, opened 2019-02-25
getAvailablePurchases does not return consistent result


wasedaigo:
### Version of react-native-iap
2.2.0

### Version of react-native
0.57.4

### Platforms you faced the error (IOS or Android or both?)
iOS

### Expected behavior
getAvailablePurchases() should return consistent result

### Actual behavior
getAvailablePurchases() does not return correct values if it is called without much interval

### Tested environment (Emulator? Real Device?)
Real Device (probably only sandbox...? It is  hard to test in production)

### Steps to reproduce the behavior
Call getAvailablePurchases() as follows.

If I call as the follows the result can be very random.(The first time is always correct. I need to wait  for enough time then it returns 7 again)

```
      purchases = await RNIap.getAvailablePurchases();
      console.log('restorepurchase1', purchases.length);  // 7
      purchases = await RNIap.getAvailablePurchases();
      console.log('restorepurchase2', purchases.length);  // 2
      purchases = await RNIap.getAvailablePurchases(); 
      console.log('restorepurchase3', purchases.length);  // 1
      purchases = await RNIap.getAvailablePurchases();
      console.log('restorepurchase4', purchases.length);  // 2
      purchases = await RNIap.getAvailablePurchases();
      console.log('restorepurchase5', purchases.length)  ;// 2
```

hyochan:
We are having a few issues in `restoring` purchases in `ios`. We need the help of the community to find out the problem and willing to providing a solution.

wasedaigo:
I think this is sandbox only issue, it did not seem to occur in production. In that case... this bug does nothing to do with this library?

wasedaigo:
Closing this as this is sandbox related issue, which is apple's fault.

PR #436, opened 2019-03-25
Added purchased state to the restored items in getAvailableItems for iOS

backmeupplz:
Hi there! This should probably address https://github.com/dooboolab/react-native-iap/issues/409. From my experience, apple iaps api sometimes has difficulty distinguishing between `purchased` and `restored` state. Looks like when available items are returned as `restored` once — some of them will be returned as `purchased` the next time.


backmeupplz:
My PR has been up for 12 minutes and still not merged into master, the release is still not cut, I'm outraged

backmeupplz:
Actually, never mind. `Purchased` returns all the purchases that happened

PR #455, opened 2019-04-12
Support to iOS subscription offers

renatoalencar:
iOS launched this new feature that allows apps to have offers for renewing users, in order to "win back lapsed subscribers or retaining current subscribers", says the documentation.

So, this PR has two main components:

### The discounts field in products field

The `SKProduct` now has a `discounts` property in order to provide the actual subscription available discounts, I add a new method to parse it and convert to an `NSArray<NSDictionary *>` to be converted to JSON.

### The buyProductWithOffer method

This new method allows the user to purchase an offered subscription, the Apple documentation says you must generate a signature to determine the eligibility of the user to this offer, you must do it on your server using a pre-generated private key.

> Your server should respond with the signature string and the additional values it used to generate the signature: a nonce, timestamp, and the keyIdentifier. Use these values to complete the parameters in the SKPaymentDiscount object nested in the SKMutablePayment object representing the offer. If the App Store determines that the signature is not a match for the parameters in the payment, the transaction fails.

You must you `usernameHash` in the signature and transaction to identify the user, it can anything that identifies the user (an ID, UUID, username, etc).

Finally, you pass all these stuff to the `buyProductWithOffer`, like:

```js
RNIap.buyProductWithOffer(sku, user, {
  identifier: "00000...",
  keyIdentifier: "<key generated on your server>",
  none: "c2ce39dc-5d6f-11e9-8647-d663bd873d93", // The nonce is UUID and looks like this, also generated on you server
  signature: "<signature generated on your server>",
  timestamp: 1555107148.024093
})
  .then(() => console.log('the transaction was queued'))
  .catch(() => console.error('something went wrong'));
```

### Further information:

* https://developer.apple.com/documentation/storekit/in-app_purchase/implementing_subscription_offers_in_your_app?language=objc
* https://developer.apple.com/documentation/storekit/skpaymentdiscount?language=objc
* https://developer.apple.com/documentation/storekit/skproductdiscount?language=objc
* https://developer.apple.com/documentation/storekit/skproduct/3043534-discounts?language=objc

I would be grateful for any feedback, thanks.

:smile:

### DIFF
--- a/index.js
+++ b/index.js
@@ -318,7 +318,7 @@ export const buyPromotedProduct = () => Platform.select({
  * @param {number} withOffer.timestamp The timestamp of the signature
  * @returns {Promise}
  */
-export const buyProductWithOffer = (sku, forUser, withOffer) => Platform.select({
+export const buyProductWithOfferIOS = (sku, forUser, withOffer) => Platform.select({
   ios: () => {
     checkNativeiOSAvailable();
     return RNIapIos.buyProductWithOffer(sku, forUser, withOffer);
--- a/index.js
+++ b/index.js
@@ -305,7 +305,7 @@ export const buyPromotedProduct = () => Platform.select({
 
 /**
  * Buy products or subscriptions with offers (iOS only)
- * 
+ *
  * Runs the payment process with some infor you must fetch
  * from your server.
  * @param {string} sku The product identifier
--- a/index.js
+++ b/index.js
@@ -304,6 +304,29 @@ export const buyPromotedProduct = () => Platform.select({
 })();
 
 /**
+ * Buy products or subscriptions with offers (iOS only)
+ * 
+ * Runs the payment process with some infor you must fetch
+ * from your server.
+ * @param {string} sku The product identifier
+ * @param {string} forUser  An user identifier on you system
+ * @param {object} withOffer The offer information
+ * @param {string} withOffer.identifier The offer identifier
+ * @param {string} withOffer.keyIdentifier Key identifier that it uses to generate the signature
+ * @param {string} withOffer.nonce An UUID returned from the server
+ * @param {string} withOffer.signature The actual signature returned from the server
+ * @param {number} withOffer.timestamp The timestamp of the signature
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -303,6 +303,29 @@ export const buyPromotedProduct = () => Platform.select({
...
+ * @param {string} forUser  An user identifier on you system
+ * @param {object} withOffer The offer information
+ * @param {string} withOffer.identifier The offer identifier
+ * @param {string} withOffer.keyIdentifier Key identifier that it uses to generate the signature
+ * @param {string} withOffer.nonce An UUID returned from the server
+ * @param {string} withOffer.signature The actual signature returned from the server
+ * @param {number} withOffer.timestamp The timestamp of the signature
+ * @returns {Promise}
+ */
+export const buyProductWithOffer = (sku, forUser, withOffer) => Platform.select({
```
hyochan: What about renaming this to `buyProductWithOfferIOS`?

renatoalencar: Right, I'll do this.

renatoalencar: Done

ios/RNIapIos.m
```diff
@@ -463,12 +501,87 @@ -(NSDictionary*)getProductObject:(SKProduct *)product {
...
+          break;
+        default:
+          discountType = @"";
+          break;
+      }
+
+      [mappedDiscounts addObject:[NSDictionary dictionaryWithObjectsAndKeys:
+        discount.identifier, @"identifier",
+        discountType, @"type",
+        [@(discount.numberOfPeriods) stringValue], @"numberOfPeriods",
```
sck-v: Please check https://github.com/dooboolab/react-native-iap/pull/454

I think there can be the same issue

renatoalencar: Ok, will change it soon.

renatoalencar:
Right @hyochan, I will add TS and Flow typings.

renatoalencar:
Done

kristfal:
Any reason why this is not merged yet?

PR #459, opened 2019-04-20
iOS: Return bool in canMakePayments

raeesbhatti:
[The API](https://github.com/dooboolab/react-native-iap#methods) defines that `initConnection` returns a string in promise. But on Android, it was returning a boolean. This change fixes that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Also, note that this is our last migration for renaming method names without any
 | Func  | Param  | Return | Description |
 | :------------ |:---------------:| :---------------:| :-----|
 | ~~prepare~~ |  | `Promise<void>` | Deprecated. Use `initConnection` instead. |
-| initConnection |  | `Promise<string>` | Init IAP module. On Android this can be called to preload the connection to Play Services. In iOS, it will simply call `canMakePayments` method and return value.|
+| initConnection |  | `Promise<boolean>` | Init IAP module. On Android this can be called to preload the connection to Play Services. In iOS, it will simply call `canMakePayments` method and return value.|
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/dooboolab/RNIap/RNIapModule.java
```diff
@@ -158,7 +158,7 @@ public void initConnection(final Promise promise) {
     ensureConnection(promise, new Runnable() {
       @Override
       public void run() {
-        promise.resolve(true);
+        promise.resolve("true");
```
hyochan: Could you change the string into something readable? Something like `android billing connected`?

raeesbhatti: I used `"true"` because iOS returns the same. Are you sure about this?

raeesbhatti:
> @raeesbhatti Thank you for the contribution. Please see my review.

@hyochan I used "true" because iOS returns the same. Are you sure about this?

android/src/main/java/com/dooboolab/RNIap/RNIapModule.java
hyochan: Oh. Then it would be better for ios to give better string too.  Something like `Can make payment in ios.`. Since `true` is very useless.

raeesbhatti: Since the promise will only be resolved if `ensureConnection` succeeds and there is only one outcome. It think, it would be better to return a Boolean `true` here, and in iOS also. Or just make the function void. This would make it easier to consume the outcome in JS, because it's much easier to check for Boolean than to compare a string. A string message would make sense if there were multiple successful outcomes.

PR #472, opened 2019-04-30
Remove comments about ejected apps in README

emilioicai:
Removing `pod 'RNIap', :path => '../node_modules/react-native-iap'` from the Podfile makes the module not available in apps using Pods so it's better to let it be in the Podfile.


hyochan:
I am not quite sure about this, but in some `env` like in `expokit` that actually worked out for `linking` correctly. I'll try to merge this after few days.

Issue #423, opened 2019-03-14
Better `StoreKit` flow.


hyochan:
Moving discussion from https://github.com/dooboolab/react-native-iap/issues/307#issuecomment-467916155.

We are willing to handle better store kit flow.

hyochan:
@Edgpaez I think I can just remove `RNIap.addAdditionalSuccessPurchaseListenerIOS` which comes after failure just by encapsulating. Followed by [PR](RNIap.addAdditionalSuccessPurchaseListenerIOS) I've made in `flutter_inapp_purchase`.

hyochan:
@Edgpaez Actually you were right. Today, I just reviewed `flutter_inapp_purchase` and tried to integrate to `react-native-iap`, and realized that I was using the `stream` in dart so I didn't have to expose `addAdditionalSuccessPurchaseIOS` listener to production code.
Therefore, your idea of using `RxJS` would be really great to have it here.

cayasso:
Hi guys, thanks for this great discussion? any update on this, we are experiencing the same issue mentioned in https://github.com/dooboolab/react-native-iap/issues/307 pretty bad user experience, I would contribute but have no experience in ios native language whatsoever :(

hyochan:
@cayasso Currently, this discussion is about `javascript` side and not related to `native part`. Originally @Edgpaez issued that we can encapsulate `addAdditionalSuccessPurchaseListenerIOS` in `index.js` and not expose to actual usage so the user can end up with `buyProduct` method without calling extra listener. This could be done using `rxjs`. If there is no contribution to this for a while, I'll have to think on different structures.

Edgpaez:
Hi @hyochan, I'll be happy to take a look. I think your proposal:

```javascript
RNIap.buyProduct('com.example.coins100');
RNIap.buyProduct('com.example.coins200');

// receiving events
const subs =  RNIap.purchaseStatusListener(purchase => {
  ...
});
```

might be easier to reason about without prior experience in Rx. So I'll try to implement it. AFAICT, we only need JS changes since `purchase` has a `productId` property that would let users figure out which product was "purchased".


@cayasso: you can use the flow described [here](https://github.com/dooboolab/react-native-iap#purchase). No need to wait for the change in the API

ChrisEelmaa:
[...]

PR #486, opened 2019-05-13
update readme 

Edgpaez:
related to #423

`addAdditionalSuccessPurchaseListenerIOS` needs to be called prior to `buyProduct` if one expects it to catch successful purchases. 

This PR reflects this on the example usage of `buyProduct`

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,20 +189,20 @@ componentWillUnmount() {
 Once you have called `getProducts()`, and you have a valid response, you can call `buyProduct()`. Subscribable products can be purchased just like consumable products and users can cancel subscriptions by using the iOS System Settings.
 
 ```javascript
-  try {
-    // Will return a purchase object with a receipt which can be used to validate on your server.
-    const purchase = await RNIap.buyProduct('com.example.coins100');
-    this.setState({
-      receipt: purchase.transactionReceipt, // save the receipt if you need it, whether locally, or to your server.
-    });
-  } catch(err) {
-    // standardized err.code and err.message available
-    console.warn(err.code, err.message);
-    const subscription = RNIap.addAdditionalSuccessPurchaseListenerIOS(async (purchase) => {
-      this.setState({ receipt: purchase.transactionReceipt }, () => this.goToNext());
-      subscription.remove();
-    });
-  }
+try {
+  // add additional iOS listener before purchasing
+  const subscription = RNIap.addAdditionalSuccessPurchaseListenerIOS(async (purchase) => {
+    this.handlePurchaseSuccess(purchase);
+    subscription.remove();
+  });
+
+  // Will return a purchase object with a receipt which can be used to validate on your server.
+  const purchase = await RNIap.buyProduct('com.example.coins100');
+  this.handlePurchaseSuccess(purchase);
+} catch(err) {
+  // standardized err.code and err.message available
+  console.warn(err.code, err.message);
+}
 ```
 
 Most likely, you'll want to handle the 'store kit flow' (detailed [here](https://forums.developer.apple.com/thread/6431#14831)), which happens when a user succesfully pays after solving a problem with his or her account - for example, when the credit card information has expired. 


vafada:
I haven't tested this but it looks like for a normal flow (not the StokeKit flow), `this.handlePurchaseSuccess(purchase);` will be called twice

hyochan:
@Edgpaez I partially agree with @vafada. You should only subscribe to the listener when there was a failure in the `purchase` unless you should put any other condition inside the listener that would avoid recording double purchases.

In addition, the `react-native-iap` won't affect on `duplicate` purchase just by adding `listener`. The side effect on doing this is receiving the same result twice. This may cause duplication when recording successful result in your database (if you are doing so).

Edgpaez:
@vafada, yes. This is not related to storekit flow. TBH, I don't understand how to address your comment. 

@hyochan I don't think you can subscribe AFTER you get the error. Not if you want to be sure you'll catch the error. 
Can we be sure the second `purchaseProcess` is called after the user caught the error and called `addAdditionalSuccessPurchaseListenerIOS` ?

I think we need to be clear to users that `this.handlePurchaseSuccess(purchase);` can be called more than once.
What do you think?

Edgpaez:
What I'm trying to do is fix the documentation since I believe the example code has a bug.

The bug (I believe) is that `addAdditionalSuccessPurchaseListenerIOS` won't catch all success purchases in the storekit flow.

**Current Implementation**
Internally `addAdditionalSuccessPurchaseListenerIOS` relies on `NativeEventEmitter` to listen for the `iap-purchase-event` event.

Inside `RNIapIos` (the class handling the purchases natively) when the `iap-purchase-event` event [is emitted](https://github.com/dooboolab/react-native-iap/blob/5e7257f45a2375b7d47f4b5e84b3abf465740336/ios/RNIapIos.m#L377) there is no guarantee that `addAdditionalSuccessPurchaseListenerIOS` has already registered to listen that event.

**So, what's the problem?**
I think that means that there is no way to be sure that when the `iap-purchase-event` is emitted natively, there is a listener registered.
That means there is a possibility that users successfully pay but JS doesn't acknowledge  the payment.

**The fix?**
Registering the listener for the `iap-purchase-event` event before making a purchase.


I hope that's clearer @hyochan

hyochan:
@Edgpaez Thanks for your through explanation and I could understand you much better! I've not yet reproduced any missing event in my other app. Also, the listener is registered when there was failure in the purchase as in [line](https://github.com/dooboolab/react-native-iap/blob/2b355f7d66d8612e920a3dfcf4b67657a00cdb40/index.js#L387). I am not quite sure tough whether there will be any hole doing the implementation this way since this was just the work around described in #348. If there is any clear reproducible senario, I would like to try to consider this again.

vafada:
@hyochan 

@Edgpaez has a point. There is **NO GUARANTEE** that the listener would be added before the second success transaction comes to the queue. 

So this flow is possible:

1. Failed transaction enters the Payment Queue
2. Success transaction enters the Payment Queue
3. code goes to `catch` block
4. listener gets added

At this point, the success transaction is now lost.

hyochan:
@vafada The code goes to `catch` block right after `failed transaction` and starts the listener. Then the `success` transaction enters the payment queue and listener will receive it.

If we use the updated code, the listener will receive the `success` queue at the same time with the `result` of `buyProduct`. We are getting two results for `ios` for every success results which we need more codes to prevent receiving the same results and doing things twice.

vafada:
@hyochan 

We don't know how fast or slow StoreKit will put the success transaction in the queue. 

It's all async so it is not safe to assume that the success transaction will always come in **after** adding the listener.

I agree also with the listener getting two event for the same successful transaction.

This is not an easy problem :smile: 

I'm now using this fork https://github.com/superandrew213/react-native-in-app-utils/tree/listen-for-purchase-event#helpers

And here's my pseudo code to deal with StoreKit flow:

https://github.com/chirag04/react-native-in-app-utils/issues/202#issuecomment-492618182

Edgpaez:
@hyochan IMHO it's better if users handle purchases more than once instead of having **no** guarantee that _all_ successful purchases are handled. Since transactions have IDs users have data available to avoid duplicates on their end.

I know it's not ideal but IMHO it's the best suggestion we can give users until we can come up with a better API that hides all these details.


@vafada I'll take a look at your fork 👍

hyochan:
@Edgpaez If you'd like to support a better approach to the `readme` please do not erase the previous one and add to another `paragraph` for your approach and add the description on why it'd be better.
Since current approaches are not either `perfect` it'd be better to have more references.

Thanks for all your supports in advance.

hyochan:
@vafada Since `version 3` is out today. This doc is no longer needed. I may go to deprecated readme.

PR #521, opened 2019-06-14
Improved handling of receipts

emilioicai:
We noticed obtaining the App receipt isn’t as straightforward as we thought. 

NSBundle.mainBundle().appStoreReceiptURL returns an NSURL describing the path behind which the receipt can be found, if there is one!

To quote the documentation: (https://developer.apple.com/documentation/foundation/nsbundle/1407276-appstorereceipturl)

For an application purchased from the App Store, call this method on its application bundle locate the receipt. This method makes no guarantee about whether there is a file at the returned URL — **only that if a receipt is present, that is its location.**

In such a case, you have to start a SKReceiptRefreshRequest. This tries to download a receipt and saves it in the known destination if everything succeeds.

Obviously the process must be asynchronous now.

On top fo this, we fixed a couple of minor issues with v3.0.0 and added a new method to request the receipt in iOS whenever it's needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@
 | consumeAllItemsAndroid | | `Promise<void>` | Consume all items in android so they are able to buy again (on Android.) No-op on iOS. |
 | validateReceiptIos | `object` receiptBody, `boolean` isTest | `object or boolean` result | validate receipt for ios. |
 | validateReceiptAndroid | `string` packageName, `string` productId, `string` productToken, `string` accessToken, `boolean` isSubscription | `object or boolean` result | validate receipt for android. |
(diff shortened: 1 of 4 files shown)


hyochan:
@emilioicai Really thanks for this! Could you kindly rebase this with`master` once again? There seems to remain conflict rebasing it with the `master` branch.

emilioicai:
Hey @hyochan I tried rebasing again but it looks the fork is totally up to date. Could you point out which commit is missing?

hyochan:
image
I am still seeing above. Did you `fetched` upstream?

PR #529, opened 2019-06-16
fixed typo in TypeScript definition file

valdisz:
There was no  `purchaseErrorListener` defined and `purchaseUpdatedListener` was added twice.


hyochan:
I think there was error in merging Thanks.

PR #533, opened 2019-06-18
Updated method name in readme

thekevinbrown:
`endConnection()` no longer exists on this package, the function is now called `endConnectionAndroid()`. This change updates the name of the method in the docs.


bradleyflood:
Just came here to do the same :)

PR #594, opened 2019-07-17
Fix flow type for purchaseUpdatedListener

mars-lan:


hyochan:
@mars-lan Hey! Thanks again for fixing the `flow` type again. Sorry that I'm always missing things in typing.

Issue #606, opened 2019-07-22
NullPointerException when calling billingClient


mars-lan:
### Version of react-native-iap
3.3.5

### Version of react-native
0.60.4

### Platforms you faced the error (IOS or Android or both?)
Android

### Expected behavior
Not crash

### Actual behavior
Crashing with this stack trace

java.lang.NullPointerException: 
  at com.dooboolab.RNIap.RNIapModule$5.run (RNIapModule.java:229)
  at com.dooboolab.RNIap.RNIapModule$2.onBillingSetupFinished (RNIapModule.java:102)
  at com.android.billingclient.api.BillingClientImpl$BillingServiceConnection$1.run (BillingClientImpl.java:1518)
  at android.os.Handler.handleCallback (Handler.java:789)
  at android.os.Handler.dispatchMessage (Handler.java:98)
  at android.os.Looper.loop (Looper.java:180)
  at android.app.ActivityThread.main (ActivityThread.java:6950)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:240)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:835)

### Tested environment (Emulator? Real Device?)
Real Devices

### Steps to reproduce the behavior
This happens sporadically, most likely due to timing issue.

PR #607, opened 2019-07-22
Check billingClient before invoking callback

mars-lan:
This fixes the case when onHostDetroy is invoked before onBillingSetupFinished, leading to a NPE in the callback (https://github.com/dooboolab/react-native-iap/issues/606)


hyochan:
Thanks for the contribution!

Issue #627, opened 2019-08-07
'purchase-updated' is not supported event type for RNIapIos


Bang9:
### Version of react-native-iap
3.3.8

### Version of react-native
0.60.4

### Platforms you faced the error (IOS or Android or both?)
iOS only
### Expected behavior
add purchaseUpdatedListener

### Actual behavior
exception 'purchase-updated' is not supported event

### Tested environment (Emulator? Real Device?)
real device

I checked 'purchase-updated'  is already listed on supportedEvents

![image](https://user-images.githubusercontent.com/26326015/62595571-d6c3fc00-b919-11e9-8360-9778e6f0e199.png)

So I tried below, but not working
```
remove ios/Pods and ios/Podfile.lock
pod install
xcode clean and build
yarn start --reset-cache
```

### Steps to reproduce the behavior
Run this code on iOS
```
let purchaseUpdateSubscription = null;
purchaseUpdateSubscription = RNIap.purchaseUpdatedListener(async (purchase) => { ... });
```

![image](https://user-images.githubusercontent.com/26326015/62592478-3e744a00-b90e-11e9-8887-b518e59f8c1c.png)

hyochan:
It feels like you’ve not properly linked the updated module

Bang9:
> It feels like you’ve not properly linked the updated module
@hyochan Sorry I feel same, I checked now current example project working properly
WIP migrated from 2.5.5, I'll share the solution when I resolved this issue
thanks

Bang9:
I tried to resolve this issue like below

```
1. remove node_modules, remove pods, clear all cache, reinstall all dependencies
-> not linked properly

2. remove react-native-iap, step1, install app, checking remove dependencies properly,
remove app, install react-native-iap, clear build, install pods
-> not linked properly

3. step 1, added 'react-native-iap' to react-native.configs.js with null option and linking manually
-> not linked properly

4. remove repo, clone repo
-> working ¯\_(ツ)_/¯
```

PR #658, opened 2019-08-21
chore: fix podspec

Bang9:
fix source_files path and tag version format
maybe related with #627, but I'm not sure

### DIFF
--- a/RNIap.podspec
+++ b/RNIap.podspec
@@ -4,14 +4,13 @@ package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
 Pod::Spec.new do |s|
   s.name         = "RNIap"
   s.version      = package['version']
-  s.summary      = "RNIap"
-  s.description  = package['description']
+  s.summary      = package['description']
   s.homepage     = "https://github.com/dooboolab/react-native-iap"
   s.license      = "MIT"
   s.author       = package['author']
-  s.platform     = :ios, "7.0"
-  s.source       = { :git => "https://github.com/dooboolab/react-native-iap", :tag => "v#{s.version}" }
-  s.source_files  = "ios/**/*.{h,m}"
+  s.platform     = :ios, "9.0"
+  s.source       = { :git => "https://github.com/dooboolab/react-native-iap.git", :tag => "#{s.version}" }
+  s.source_files = "ios/*.{h,m}"
   s.requires_arc = true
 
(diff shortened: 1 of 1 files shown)


RNIap.podspec
```diff
@@ -4,14 +4,13 @@ package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
...
-  s.description  = package['description']
+  s.summary	 = package['description']
   s.homepage     = "https://github.com/dooboolab/react-native-iap"
   s.license      = "MIT"
   s.author       = package['author']
-  s.platform     = :ios, "7.0"
-  s.source       = { :git => "https://github.com/dooboolab/react-native-iap", :tag => "v#{s.version}" }
-  s.source_files  = "ios/**/*.{h,m}"
+  s.platform     = :ios, "9.0"
+  s.source       = { :git => "https://github.com/dooboolab/react-native-iap", :tag => "#{s.version}" }
```
hyochan: Could you explain why you removed `v` before `#{s.version}`?

RNIap.podspec
```diff
@@ -4,14 +4,13 @@ package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
...
+  s.source_files  = "ios/*.{h,m}"
```
hyochan: Could you also explain how this could fix your problem?

RNIap.podspec
Bang9: image

This repository using version format as like '3.x.x' not 'v3.x.x'

RNIap.podspec
Bang9: It is related auto linking
current podspec source_files path is "ios/**/*.{h,m}", but there is no files to link
```
ios/RNIapIos.m
ios/RNIapIos.h
ios/IAPPromotionObserver.h
ios/IAPPromotionObserver.m
= ios/*.{h,m}
```

RNIap.podspec
hyochan: Thanks for the detail. I wasn't taking care so much on the release tags. :sweat_smile:

RNIap.podspec
hyochan: Thanks again for this! Looks quite reasonable and I'll merge it for now.

Bang9: thanks for confirming!

PR #680, opened 2019-08-31
Destroying listeners in proper React's life cycle method

404-html:
I believe that listener's `remove()` should be called in `componentWillUnmount()` instead of `componentWillMount()`.


jvandenaardweg:
That's correct

PR #691, opened 2019-09-03
Added a couple of commands on the readme that were necessary to get it working.

felipegouveiae:


jvandenaardweg:
Thanks for the suggestion!

Maybe good to add a line of text explaining that's only needed when you use Pod's in your project, or if you use React Native 0.60 and newer.

Also, can you make sure the commands are on a new line?

Screenshot 2019-09-04 at 09 59 18

README.md
```diff
@@ -144,6 +144,8 @@ Getting Started
 
 ### Mostly automatic installation
 `$ react-native link react-native-iap`
+`$ cd ios`
+`$ pod install`
```
hyochan: You should mention this for `pod user` or `react-native >= 0.60` user.

hyochan:
The current `readme` covered this `PR` in a much better way. Hope you can follow that.

PR #709, opened 2019-09-11
Fix bad return type of promise of 'initConnection'.

Johny8811:


hyochan:
Closes via #713 .
Added @Johny8811 as a `co-author`.

PR #713, opened 2019-09-15
Update package to 3.4.7

hyochan:
- Add missing `isAcknowledgedAndroid` variable in getAvailableItems.

Closes #709 
Co-authored-by: Johny8811 <jan.hujsa@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
     - Support installation guide for both `RN >= 0.60` and `RN <0.60`. No need to use different version of `react-native-iap` now.
   - **[3.4.6]**
     - Reject promises on `initConnection` when not resolved [#704](https://github.com/dooboolab/react-native-iap/pull/696).
+  - **[3.4.7]**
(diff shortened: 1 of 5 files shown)


jvandenaardweg:
@hyochan does that mean the acknowledge flow in the example is not up to date anymore for subscriptions? https://github.com/dooboolab/react-native-iap/blob/ec53ac446b5ebdd459a53dde6a52b49d089ab292/IapExample/App.js#L66

Should it be like this?

```javascript
if (purchase.purchaseStateAndroid === 1) {
    try {
        const ackResult = await acknowledgePurchaseAndroid(purchase.purchaseToken);
        console.log('ackResult', ackResult);
    } catch (ackErr) {
        console.warn('ackErr', ackErr);
    }
}
```

Also, is `purchaseStateAndroid` the same as `paymentState` in the [Android API docs](https://developers.google.com/android-publisher/api-ref/purchases/subscriptions)? The docs of this package could be updated with a little bit more info for what `purchaseStateAndroid` actually is and what possible values it could be.

PR #718, opened 2019-09-17
feat: expose error codes

doomsower:
This PR exposes all constants from Android and iOS (plus E_IAP_NOT_AVAILABLE) to JS code, for more convenient error handling.


doomsower:
I have zero experience with flow. I've added some, but I'm not sure if this is correct way.

PR #724, opened 2019-09-21
fix: improve typings

doomsower:
Sunce `buyProduct` and `requestPurchase` wrap the same native calls, they should resolve with same type of promise.

And subscriptions are now strongly typed.


hyochan:
@doomsower Sorry I've not meant to close this but just giving an opinion. As I've written in `issue`, I'll consider this when I am available.

PR #616, opened 2019-07-29
Supported old error codes

Rinkuk1993:
Fixes issue : 
https://github.com/dooboolab/react-native-iap/issues/583

Did not check iOS completely due to unavailability of account. Please let me know how I can help.

PR #738, opened 2019-09-29
Add missing fields to PurchaseResult & PurchaseError

mars-lan:
These were added as part of https://github.com/dooboolab/react-native-iap/pull/616 but the flow file never got updated


hyochan:
Thanks for taking care of this again! I am thinking of automatic conversion from `ts-to-flow` conversion using [flowgen](https://github.com/joarwilk/flowgen). Hope that can help these problems in future.

PR #806, opened 2019-10-24
[Android] Fixed promise never returned from requestSubscription()

rewieer:
Calling `requestSubscription()` on Android result in the purchase being made but the promise is never called. This is due to an exception being thrown when trying to resolve the promise, specifically `method threw 'com.facebook.react.bridge.objectalreadyconsumedexception' exception` because the WriteableMap is already consumed when `sendEvent` is called. 
By making a copy before calling it, we get a fresh WriteableMap that we can send to our promise without error.

### DIFF
--- a/android/src/main/java/com/dooboolab/RNIap/RNIapModule.java
+++ b/android/src/main/java/com/dooboolab/RNIap/RNIapModule.java
@@ -522,8 +522,8 @@ public class RNIapModule extends ReactContextBaseJavaModule implements Purchases
         item.putBoolean("isAcknowledgedAndroid", purchase.isAcknowledged());
         item.putInt("purchaseStateAndroid", purchase.getPurchaseState());
 
+        promiseItem = item.copy();
         sendEvent(reactContext, "purchase-updated", item);
-        promiseItem = item;
       }
       if (purchases.size() > 0 && promiseItem != null) {
         DoobooUtils.getInstance().resolvePromisesForKey(PROMISE_BUY_ITEM, promiseItem);


android/src/main/java/com/dooboolab/RNIap/RNIapModule.java
```diff
@@ -522,8 +522,8 @@ public void onPurchasesUpdated(BillingResult billingResult, @Nullable List<Purch
         item.putBoolean("isAcknowledgedAndroid", purchase.isAcknowledged());
         item.putInt("purchaseStateAndroid", purchase.getPurchaseState());
 
+        promiseItem = item.copy();
```
mars-lan: the `copy` method is only added in RN 0..61. Should probably bump up the major version and update doc to reflect that.

hyochan: @mars-lan Hey! Thank you for the detail. Um... I should have to think about it since I feel kind of reluctant to separate the major version. @rewieer How do you think?

rewieer: The copy method basically calls merge, so we can create a `public WriteableMap copy(WritableMap original)`.

hyochan: @rewieer Sounds great! Would you like to give a `PR` for your idea?

rewieer: @hyochan sorry for the long delay, I was about to do it but it seems someone did it before me

PR #807, opened 2019-10-24
[Android] added subs to the list of unconsumed purchases

rewieer:
`sendUnconsumedPurchases` only send unconsumed in-app products but not subscriptions. The app i'm working on uses subscriptions and requires to be able to send receipt to the server about unacknowledged subscriptions. This fixes it.

I also saw that a promise is passed as a parameter but isn't subscribed to inside `index.ts`. As is, this promise is useless so I decided to resolve it with true no matter what at the end of the function.


hyochan:
@emilioicai Could you kindly help reviewing this?

PR #858, opened 2019-11-20
Pass type of purchase in the response

varungupta85:
Pass the type of purchase such that we can find if the purchase is an in-app product or a subscription.


hyochan:
I think we need to think further for exposing `type` like this. I'll come back if I am ready.

hyochan:
I don't this is necessary since we know the purchases type when we are requesting.

Issue #801, opened 2019-10-19
purchaseUpdateListener is getting called multiple times in IOS


anuragdwivedi29:
### Version of react-native-iap
4.0.4
### Version of react-native
0.60.5
### Platforms you faced the error (IOS or Android or both?)
IOS
### Expected behavior
should be called when requestSubscription(subscriptionID) is called
### Actual behavior
is getting called when the app loads for the very first time.Also when a subscription is bought it is getting called multiple times
### Tested environment (Emulator? Real Device?)
Real Device (sandbox environment)

anuragdwivedi29:
are there any pointers for the above mentioned issue?

Bang9:
maybe it would help you `clearTransactionIOS()`, but i'm not sure :p
[reference](https://developer.apple.com/library/archive/technotes/tn2387/_index.html)

anuragdwivedi29:
ok i will try that but i am  testing the in app purchase in sandbox environment .So when i am launching the app and before making any purchase it is calling the clearTransactionIOS() and is making the purchase which should not be the behaviour. Ideally only when the user makes the purchase the it should ask for the payment and make the purchase.So therefore i am not able to understand  as to what i am doing wrong here.
Any pointers would be highly appreciated.Thanks

franj0:
@anuragdwivedi29 if you did not finish transaction on app relaunch update listener will try automatically execute code inside.

Also do not forgot this.yourListener.remove() on unmount...

Maybe something of this you need. gl

mingxin-yang:
+1

nicolaswilliman:
This is happeening to me too. The difference is that I'm using the 3.2.3 version of RNIap.
I'm also removing the listener when unmounting, but the purchaseUpdateListener is getting called multiple times with differents transaction ids. Just happens in iOS.

Any ideas?

Bang9:
@nicolaswilliman @anuragdwivedi29 @mingxin-yang @franj0 
I found the reason, refer this link(https://help.apple.com/app-store-connect/#/dev7e89e149d)

one auto-renewable subscription makes 6 transactions
If you turn the app back on 20 minutes **_after subscribe a 1week_**
**_6 transactions that occurred every 3 minutes_** will enter the purchaseUpdatedListener at once

alexpchin:
@Bang9 Do you have a solution for this? I understand that the `purchaseUpdatedListener` is being called multiple times

Bang9:
[...]

PR #890, opened 2019-12-30
Documentation updates

denkeni:
Please see respective commits.

Also proposing *not* to deprecate `clearTransactionIOS`. See: https://github.com/dooboolab/react-native-iap/issues/801#issuecomment-568748742


denkeni:
@hyochan Merge conflict was resolved.

Issue #895, opened 2020-01-08
NullPointerException on acknowledgePurchase


niklaskors:
### Version of react-native-iap

3.0.3

### Version of react-native

0.59.10

### Platforms you faced the error (IOS or Android or both?)

Android

### Expected behavior

Not crash (should have a billing client)

### Actual behavior

Crashes occasionally with
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.android.billingclient.api.BillingClient.acknowledgePurchase(com.android.billingclient.api.AcknowledgePurchaseParams, com.android.billingclient.api.AcknowledgePurchaseResponseListener)' on a null object reference
       at com.dooboolab.RNIap.RNIapModule.acknowledgePurchase(RNIapModule.java:418)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
       at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:158)
       at com.facebook.react.bridge.queue.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:761)
       at android.os.Handler.dispatchMessage(Handler.java:98)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:29)
       at android.os.Looper.loop(Looper.java:156)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:232)
       at java.lang.Thread.run(Thread.java:776)
```

### Tested environment (Emulator? Real Device?)

Real Devices in production environment 

### Steps to reproduce the behavior

I was only able to reprodruce the issue by explicitly setting billingClient to null. Apparently this can be null occasionally.

PR #896, opened 2020-01-08
Ensure connection for acknowledgement call (#895)

niklaskors:
I've noticed an occasional `NullPointerException` crash (https://github.com/dooboolab/react-native-iap/issues/895) on some of the acknowledgement calls.
I'm still using the older 3.0.3 version but I've noticed that this can still be a problem in later versions as well. 
I wonder why the code inside `acknowledgePurchase` isn't wrapped with `ensureConnection` like the other functions as this would solve the case of billingClient being null? (it's also missing in `consumeProduct` btw).
I haven't tested this in production but I'm pretty sure it will fix this particular issue.


hyochan:
Fixes #895

niklaskors:
Nice, thanks for checking this so quickly. Shall I create another PR to wrap the code inside `consumeProduct` with `ensureConnection` as well? I think that function might have the same problem. 
By the way, Is it easy for you to patch `3.0.3` with this fix as well? That would be really helpful :)

hyochan:
Sure!!! Also, I don't want to do the old fork update. Why mot upgrading instead? v3 and v4 aren't very different!

niklaskors:
Was not able to upgrade yet because I'm using react-native v0.59 which doesn't support AndroidX yet which I think is necessary for v4 right?

Issue #1020, opened 2020-05-27
App Crash after bought product successfully - iOS?


anastely:
### Version of react-native-iap
4.4.9
### Version of react-native
0.61.5
### Platforms you faced the error (IOS or Android or both?)
IOS
### Expected behavior
After user Buy the product it should finish transaction then call some functionality 'redux Stuff'  
### Actual behavior
After user Buy the product the app crashed then appear this message 

> You're All set 
> Your Purchase was successful

 
### Tested environment (Emulator? Real Device?)
Real Device ( TestFlight )

### Code 
```

import * as RNIap from 'react-native-iap';

const items = Platform.select({
  ios: ['com.myapp.premium'],
});

let purchaseUpdateSubscription;
let purchaseErrorSubscription;

export default class PurchaseScreen extends Component {
  constructor() {
    super();
    this.state = {
      availablePurchases: [],
    };
  }

  componentDidMount() {
    this.getProduct();

    purchaseUpdateSubscription = RNIap.purchaseUpdatedListener(
      async purchase => {
        const receipt = purchase.transactionReceipt;
        if (receipt) {
          console.log('receipt', receipt);
          try {
            if (Platform.OS === 'ios') {
              const finishTransactionIOS = RNIap.finishTransactionIOS(
                purchase.transactionId,
              );
              Alert.alert('finishTransactionIOS', finishTransactionIOS);
            }
          } catch (ackErr) {
            Alert.alert('ackErr', ackErr);
          }
          this.setState({receipt}, () => this.goNext());
        }
      },
    );

    purchaseErrorSubscription = RNIap.purchaseErrorListener(error => {
      console.log('purchaseErrorListener', error);
      Alert.alert('purchaseErrorSubscription error:', JSON.stringify(error));
    });
  }

  goNext = () => {
    Alert.alert(
      'Get full access now | Redux Stuff:), after second you will navigate to Home Screen immediately.',
    );
    setTimeout(() => {
      this.props.navigation.navigate('BottomTabNavigator');
    }, 1500);
  };

  componentWillUnmount() {
    if (purchaseUpdateSubscription) {
      purchaseUpdateSubscription.remove();
      purchaseUpdateSubscription = null;
    }
    if (purchaseErrorSubscription) {
      purchaseErrorSubscription.remove();
      purchaseErrorSubscription = null;
    }
    RNIap.endConnection();
  }

  getProduct = async () => {
    try {
      await RNIap.initConnection();
[...]

PR #1021, opened 2020-05-28
Update README.md

anastely:
finishTransactionIOS / finishTransaction It's a Promise So it's should await it when invoked.
 
Related in this issue [#1020](https://github.com/dooboolab/react-native-iap/issues/1020)


anastely:
As you see in the issue #1020 it's not working if I don't add await, after added it the app works very well, to be honest, I don't use it without `catch` and I guess it's important to catch errors in these cases

Issue #1058, opened 2020-07-29
RNIap.requestPurchase have no resolved promise result on iOS 14.0 version


VEnom42:
### Version of react-native-iap
4.4.8, 4.5.0
### Version of react-native
0.61.5
### Platforms you faced the error (IOS or Android or both?)
IOS (14.0)
### Expected behavior
requestSubscription and requestPurchase methods should resolve promise on sucessful purchase.
### Actual behavior
`purchaseUpdatedListener` receives the result of successful purchase but `RNIap.requestPurchase(sku, false)` and `RNIap.requestSubscription(sku, false)` have no resolved promise result. Actually there is no success callback execution. It works fine in previous ios versions and in android.
### Tested environment (Emulator? Real Device?)
both
### Steps to reproduce the behavior
1. Set up purchases.
2. Get real device with ios 14.0 or emulator with working [StoreKit](https://developer.apple.com/documentation/storekit/in-app_purchase/testing_in-app_purchases_in_xcode)
3. Try to get console.log from successful purchase:
```ts
RNIap.requestPurchase(sku, false).then(result => {console.log('result', result)});
//or
RNIap.requestSubscription(sku, false).then(result => {console.log('result', result)});
```

MacroBet:
I'm having the same problem

captaincole:
Also seeing this issue on iOS 14 beta 2.

corasan:
Weird thing is that it was working for me perfectly fine.

lempa:
We're having the same issue  in a mac catalyst app on iOS 13 / macOS 10.15.5.
The steps to reproduce the behavior are the same

iaphub:
Just submitted a pull request to fix the issue.
As soon as it is accepted, we will update [react-native-iaphub](https://github.com/iaphub/react-native-iaphub) to the latest version of react-native-iap with the fix 👍

VEnom42:
Fixed in `4.5.2`.

n0umankhan:
+1 still having the same problem in IOS, I'm using 4.5.3

PR #1064, opened 2020-08-03
Fix promise not resolving on IOS14

iaphub:
Hey @hyochan,

We fixed an issue causing the `requestSubscription` or `requestPurchase` promise to not resolve on IOS 14 (Issue: #1058)
The issue is due to how react-native-iap is handling the promises, currently a key is generated (using RCTKeyForInstance) which is the pointer of the product identifier.
The problem is the pointer of `transaction.payment.productIdentifier` is actually different in the `updatedTransactions` method on IOS 14 causing the key to identify the promise to be different therefore the promise isn't found.
The solution in the pull request is to use directly the product identifier instead of the pointer value.

### DIFF
--- a/ios/RNIapIos.m
+++ b/ios/RNIapIos.m
@@ -155,8 +155,7 @@ RCT_EXPORT_METHOD(buyProduct:(NSString*)sku
         }
     }
     if (product) {
-        NSString *key = RCTKeyForInstance(product.productIdentifier);
-        [self addPromiseForKey:key resolve:resolve reject:reject];
+        [self addPromiseForKey:product.productIdentifier resolve:resolve reject:reject];
             
         SKMutablePayment *payment = [SKMutablePayment paymentWithProduct:product];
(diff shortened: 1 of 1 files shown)


Hobahung:
Work in IOS 14 but error in IOS 13, 12...

iaphub:
Hey @Hobahung, just tested on IOS 12 and no issue.
What's your problem exactly?

Hobahung:
Dear @iaphub iap

This error occurs when I renew an Non-Renewing Subscription or Consumable item several times a day. However, only the error on IOS is less than 14. The RNIap.requestPurchase function still returns the result on IOS14, however it doesn't work on iOS 10 or 12. I am also investigating the cause of this problem. Issue when i click buy button in attached photo. Please see attached photo. 

121461327_681622175804774_8800397962450725145_n

trinhvanminh2009:
I upgraded to the latest version 4.6.3 and it still not work on requestPurchase vs getSubscriptions

Issue #1012, opened 2020-05-20
Purchase with "Slow card, always denied" is never set as canceled and always fire purchaseUpdatedListener


robinalexandre:
### Version of react-native-iap
4.4.3

### Version of react-native
0.62.2

### Platforms you faced the error (IOS or Android or both?)
Android (only tested on it for now)

### Expected behavior
For "slow card, always denied", an update should be received in `purchaseUpdatedListener` or in `purchaseUpdatedError` when the payment has been canceled. Moreover, if I relaunch the app, the listener should not be fired or at least with the real status which should be "canceled" and not "pending". 

### Actual behavior
When on testing environment, if I choose to pay with "slow card, always accepted", I receive the event in `purchaseUpdatedListener` when the payment has been accepted. However, for "slow card, always denied", there is no update receive in `purchaseUpdatedListener` when the payment has been canceled. Moreover, if I relaunch the app, the listener is fired with the same purchase and with the "pending status" whereas there is no reason to fire it because the purchase is now canceled and the status is not the good one. 

### Tested environment (Emulator? Real Device?)
Real device

### Steps to reproduce the behavior
Make a payment with "slow card, always denied".
I am using non-consumables, I don't know if it's the same behaviour for suscriptions or consumables products.

Thanks for the lib !!!

medonagy45:
I am having the same issue

johnpan188:
I have same issue here.
Can't even purchase same item again

AmauryLiet:
For info, it seems like the issue is coming from Google billing's library

Some sources:
- Issue on android repo: https://github.com/android/play-billing-samples/issues/272
Related issues:
- https://issuetracker.google.com/issues/73982566
- https://issuetracker.google.com/issues/149668311

AmauryLiet:
From my own experience, `purchaseUpdatedListener` stops being called (or at least for a shorter time), when I make sure to call `initConnection` on app mount

Are you calling `initConnection` on your side?

bcbcbcbcbcl:
I'm having the same issue, any update on this issue? Tried to call RNIap.flushFailedPurchasesCachedAsPendingAndroid() but got error "RNIap.flushFailedPurchasesCachedAsPendingAndroid is not a function".

AmauryLiet:
As seen in #1140 @bcbcbcbcbcl error seems to be fixed

This issue seems to be fixed by upgrading to v4.6.1


Should we close this issue @hyochan?

[...]

Issue #126, opened 2018-04-13
Doesn't let rebuy Android item on test account


ilyakar:
### Version of react-native-iap
0.3.10

### Platforms you faced the error (IOS or Android or both?)
Android

### Expected behavior
Should be able to rebuy an item as many times as I need since I'm just testing

### Actual behavior
Returns
```
Error: You already own this item.
    at createErrorFromErrorData (59522b9c-ca17-4e00-9e00-05cc989b1199:2016)
    at 59522b9c-ca17-4e00-9e00-05cc989b1199:1968
    at MessageQueue.__invokeCallback (59522b9c-ca17-4e00-9e00-05cc989b1199:2410)
    at 59522b9c-ca17-4e00-9e00-05cc989b1199:2155
    at MessageQueue.__guardSafe (59522b9c-ca17-4e00-9e00-05cc989b1199:2323)
    at MessageQueue.invokeCallbackAndReturnFlushedQueue (59522b9c-ca17-4e00-9e00-05cc989b1199:2154)
    at t (RNDebuggerWorker.js:1)
```

### Steps to reprodue the behabior
Buy an item more than once using a test account

ilyakar:
Hi @dooboolab, it's not a consumable product.  Users have it forever. There's a way to restore the product so it's good to let users known in production that they cannot buy the same product again since they already bought it. But for testing purposes, it gets very uncomfortable to have to use a different account each time all the purchases have been bought.

ilyakar:
I don't think I had this problem in 0.2.17

dooboolab:
@ilyakar Ah.. Sorry I was confused. Actually, if it is `non-consumable` product, you can't actually rebuy it. If that was possible in `0.2.17` that must be working awkwardly.. Only consumable products are able to rebuy it. It is same in both platforms (ios and android). If user want to restore the item they have to do it optionally but as in [LINK](https://android.stackexchange.com/questions/61426/restoring-purchase-of-premium-features-after-reinstall?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa), it won't charge the user again. Therefore, if you are designing in-app products that you want to charge users by rebuying, I think you should create `consumable` product.

dooboolab:
[...]

PR #1085, opened 2020-08-26
Fix/avoid blindly consuming success purchases

AmauryLiet:
### The issue

I noticed that using `consumeAllItemsAndroid` is **pretty dangerous in non-dev** environnement:
If a purchase becomes successful while the app is offline or killed, then calling `consumeAllItemsAndroid` will consume the successful purchase without ever notifying the host app, preventing the purchased product from being delivered (although it is billed)

However `consumeAllItemsAndroid` still appears to be pretty **useful in some prod cases**:
When a pending payment fails:
- the app is never notified of the fail (#1012)
- the purchase remains pending due to Play Store cache (related Android issues: https://github.com/android/play-billing-samples/issues/272, https://issuetracker.google.com/issues/73982566 & https://issuetracker.google.com/issues/149668311)

The only way (that I am aware of) to force the cache refresh is to try to consume the pending purchase. You would then get a `BillingResponseCode.ITEM_NOT_OWNED` error, and the item is available for purchase again. (credit to [this post](https://github.com/android/play-billing-samples/issues/272#issuecomment-667688696))
This fix was already suggested in some issues: https://github.com/dooboolab/react-native-iap/issues/126#issuecomment-439084872

### The solution suggested in this PR

We should expose a method to flush all failed purchases that are still - wrongly - marked as pending.

It would simply fetch all **pending** purchases, and try to consume then individually:
- if they are indeed pending, we would get `BillingResponseCode.DEVELOPER_ERROR`
- if they were actually already failed, we would get `BillingResponseCode.ITEM_NOT_OWNED`

-----

Happy to get your feedback on the implementation/naming/...

I am not a big fan of the "callback hell" of the readme. I guess migrating the example/doc to hooks would help but that a subject for another PR

Worthy discussion:
- should that be the behaviour by default (done by default in `initConnection`)?

### DIFF
--- a/IapExample/App.js
+++ b/IapExample/App.js
@@ -116,7 +116,7 @@ class Page extends Component {
   async componentDidMount(): void {
     try {
       const result = await RNIap.initConnection();
-      await RNIap.consumeAllItemsAndroid();
+      await RNIap.flushFailedPurchasesCachedAsPendingAndroid();
       console.log('result', result);
     } catch (err) {
       console.warn(err.code, err.message);
--- a/README.md
+++ b/README.md
@@ -124,7 +124,8 @@ _*deprecated_<br>~~`buySubscription(sku: string)`~~<ul><li>sku: subscription ID/
 `getPendingPurchasesIOS()` | `Promise<ProductPurchase[]>` | **IOS only**<br>Gets all the transactions which are pending to be finished.
 `validateReceiptIos(body: Record<string, unknown>, devMode: boolean)`<ul><li>body: receiptBody</li><li>devMode: isTest</li></ul> | `Object\|boolean` | **iOS only**<br>Validate receipt.
 `endConnection()` | `Promise<void>` | End billing connection.
-`consumeAllItemsAndroid()` | `Promise<void>` | **Android only**<br>Consume all items so they are able to buy again.
+`consumeAllItemsAndroid()` | `Promise<void>` | **Android only**<br>Consume all items so they are able to buy again. ⚠️ Use in dev only (as you should deliver the purchased feature BEFORE consuming it)
+`flushFailedPurchasesCachedAsPendingAndroid()` | `Promise<void>` | **Android only**<br>Consume all 'ghost' purchases (that is, pending payment that already failed but is still marked as pending in Play Store cache)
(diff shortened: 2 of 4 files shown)


AmauryLiet:
TODO: update README

AmauryLiet:
@iaphub @hyochan do you have any feedback on the PR?
is the issue clear to you?

hyochan:
> ### The issue
> I noticed that using `consumeAllItemsAndroid` is **pretty dangerous in non-dev** environnement:
> If a purchase becomes successful while the app is offline or killed, then calling `consumeAllItemsAndroid` will consume the successful purchase without ever notifying the host app, preventing the purchased product from being delivered (although it is billed)
> 
> However `consumeAllItemsAndroid` still appears to be pretty **useful in some prod cases**:
> When a pending payment fails:
> 
> * the app is never notified of the fail (#1012)
> * the purchase remains pending due to Play Store cache (related Android issues: [android/play-billing-samples#272](https://github.com/android/play-billing-samples/issues/272), https://issuetracker.google.com/issues/73982566 & https://issuetracker.google.com/issues/149668311)
> 
> The only way (that I am aware of) to force the cache refresh is to try to consume the pending purchase. You would then get a `BillingResponseCode.ITEM_NOT_OWNED` error, and the item is available for purchase again. (credit to [this post](https://github.com/android/play-billing-samples/issues/272#issuecomment-667688696))
> This fix was already suggested in some issues: [#126 (comment)](https://github.com/dooboolab/react-native-iap/issues/126#issuecomment-439084872)
> 
> ### The solution suggested in this PR
> We should expose a method to flush all failed purchases that are still - wrongly - marked as pending.
> 
> It would simply fetch all **pending** purchases, and try to consume then individually:
> 
> * if they are indeed pending, we would get `BillingResponseCode.DEVELOPER_ERROR`
> * if they were actually already failed, we would get `BillingResponseCode.ITEM_NOT_OWNED`
> 
> Happy to get your feedback on the implementation/naming/...
> 
> I am not a big fan of the "callback hell" of the readme. I guess migrating the example/doc to hooks would help but that a subject for another PR
> 
> Worthy discussion:
> 
> * should that be the behaviour by default (done by default in `initConnection`)?

Migrating to `react-hook` is really a great idea. I am not quite sure what you mentioned on the default behavior on `initConnection` tough 🤔.

PR #1089, opened 2020-08-28
feat: Allow specyfing string subtype for product ids

AmauryLiet:
# Issue

I want to protect myself by restraining the acceptable skus in my application (either with `type ItemSku = 'sku1' | 'sku2' | ...` or `enum ItemSku { sku1 = 'sku1', sku2 = 'sku2', ... }`

However, `getProducts` returns products with `productId` typed as `string`

# Fix

getProducts now infers the type of `productId` from its input

This is backward compatible as the typing is optionnal and defaults to the current `string`

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -55,9 +55,9 @@ export interface Discount {
   subscriptionPeriod: string;
 }
 
-export interface Product extends Common {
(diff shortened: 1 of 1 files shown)


hyochan (approved):
LGTM!

PR #1100, opened 2020-09-04
feat: Let user know which productId was the error about

AmauryLiet:
Let users know which productId is the error about, on purchaseErrorListener

Always provided for ios. Not provided for android for non product-specific error


AmauryLiet:
Limitation of the feature: I could not find a way to get the sku in case of cancelled event on Android
(https://github.com/dooboolab/react-native-iap/pull/1100/files#diff-b21dbc2d57e14bf6f1b5f81c3a3c3ba5R531)

Afraid that this isn't possible natively...

hyochan (approved):
Looks great to me! However, could we also pick up which `productId` was the success about?

PR #1101, opened 2020-09-04
test: Be more specific in purchaseErrorListener param typing

AmauryLiet:
Not sure of the reason why `any` was here at the beginning. Any clue?


hyochan (approved):
I don't quite remember currently 😢 
However, looks like this is much safer. So let's take it 💯

PR #1124, opened 2020-09-22
Adding JavaVersion on build.gradle to work with Detox tests

heliojuniorkroger:
I was unable to create Detox tests without these lines of code specifying the Java version in build.gradle. Basically, without the JavaVersion, the Multidex conversion when compiling tests doesn't work well.


hyochan (approved):
Thanks for your contribution!

PR #1129, opened 2020-09-25
Fix changing subscriptions on Android

### LINKED ISSUE #1128: Changing subscriptions on Android broken
4.6.1
### Version of react-native
0.63.2
Android
### Expected behavior
When changing from one subscription to another (from yearly to monthly etc.) it should handle the old subscription properly according to the set prorationModeAndroid .
### Actual behavior
New subscription is created without anything happening to the existing subscription.
Emulator and device.
- Launch app with at least 2 subscriptions available on Google Play
- Subscribe to one of the subscriptions
- Subscribe to second subscription
- Watch both subscriptions be active at the same time

M46N3:
Added the new way to upgrade/downgrade from a previous subscription when subscribing to a new one on Android. Calls setOldSku as described in [Android docs](https://developer.android.com/reference/com/android/billingclient/api/BillingFlowParams.Builder#setOldSku(java.lang.String,%20java.lang.String))

This PR fixes #1128


hyochan (approved):
Looks like an awesome fix. However, don't you think we can rename `purchaseToken` to `purchaseTokenAndroid` if this isn't needed in `iOS` in the next updates?

M46N3:
Yes, indeed that would probably be a better name, both for future updates and stating more clearly what it does.

PR #1141, opened 2020-10-05
return ability to set 'obfuscatedAccountId' and 'obfuscatedProfileId' for android billing v3

BNTR:
Hey! I created this PR just because after I had bumped the version of lib to the new one, I noticed that the functionality of passing user id to android billing is missing. It seems that it was removed accidently and it is better to add it to fullfill API of android billing v3.
Also it seems that it related to this issue #1125

PR #1149, opened 2020-10-09
Fix amazon buyItemByType missing parameters

iaphub:
The merge of #1141 caused an issue with the amazon buyItemByType method


hyochan (approved):
Thanks for the fixes! I've also added @iaphub as a maintainer of `react-native-iap`.

PR #1182, opened 2020-11-09
Add introductory price as string without formating and currency

manishaggarwalm:
Add introductory price as string without formating and currency


hyochan (changes requested):
Thanks for your contribution. If `introductoryPriceAsAmount` is only returned by `iOS` what about changing this var to `introductoryPriceAsAmountIOS`?

manishaggarwalm:
> Thanks for your contribution. If `introductoryPriceAsAmount` is only returned by `iOS` what about changing this var to `introductoryPriceAsAmountIOS`?

Fixed the naming form `introductoryPriceAsAmount` to `introductoryPriceAsAmountIOS`.

hyochan (approved):
Thanks for the changes!! 💯

PR #1186, opened 2020-11-13
Add iOS store country code

jochem725:
This adds the App Store country code to iOS


Klaasvaak:
We would like to move from `react-native-in-app-utils` and this missing attribute is stopping us from making that move.

hyochan (approved):
Thanks for your contribution 👍

ninjaboy:
@jochem725, seems like this change is causing empty fields for products to happen randomly. I have started observing this as well as many people reported in #1203. Do you have any idea on what can be causing this?

jochem725:
@ninjaboy Yes. What happens is that countryCode somehow ends up being `nil`, everything after `nil` is not added to the NSDictionary. I created #1246 which should prevent this from happening.

PR #1199, opened 2020-12-05
Fix that getPendingPurchasesIOS() may returns undefined.

gki:
Although `getPendingPurchasesIOS()` typescript definition mentioned like below, 
```ts
/**
 * Get the pending purchases in IOS.
 * @returns {Promise<ProductPurchase[]>}
 */
export const getPendingPurchasesIOS = async (): Promise<ProductPurchase[]> => {
  if (Platform.OS === 'ios') {
    await checkNativeiOSAvailable();

    return RNIapIos.getPendingTransactions();
  }
};
```
Objective-C code had possibility to return `undefined` because it passed `nil` for `resolve()`.

(My app received a crash report because of this. 😭 )


hyochan (approved):
Great fixes! [] better than undefined

PR #1201, opened 2020-12-07
feat: added presentCodeRedemptionSheetIOS

Bang9:
This method displays a sheet that enables users to redeem subscription offer codes and works only on real-devices.

image


hyochan (approved):
OMG! You rock :rocket:

sumariana:
does anyone knows how to handle the callback? because everytime i finish the code redemption sheet, the app got restart.. the currentpurchase is not triggered

bang9:
> does anyone knows how to handle the callback? because everytime i finish the code redemption sheet, the app got restart.. the currentpurchase is not triggered

AFAIK, After redemption, a new payment transaction is delivered from SKPaymentQueue via event.

PR #1219, opened 2020-12-31
Improving typescript types for iOS subscription

alesmraz:
Improving types for iOS subscription based on current `RNIapIos.m` file


lahivee:
good idea!

hyochan (approved):
Better than string. However, could we replace the empty string with something better next time? Thanks for your contribution!

Issue #1120, opened 2020-09-20
iOS 14: requestSubscription() issues


the-dut:
### Version of react-native-iap
4.4.9

### Version of react-native
.62.2

### Platforms you faced the error (IOS or Android or both?)
iOS only.  Not testing on Android yet.

### Expected behavior
Purchase dialog pops up to complete buy and then purchaseUpdatedListener called.

### Actual behavior
call requestSubscription().  No purchase dialogs.  PurchaseUpdatedListener called immediately with no purchase dialogs or valid receipt.

### Tested environment (Emulator? Real Device?)
Real iOS 14 device using TestFlight.

### Steps to reproduce the behavior
call RNIap.requestSubscription().

Notes:
- XCode 12 
- using Sign In with Apple Id.
- testing in sandbox with real Apple id (not test user)
- works fine on iPad running iOS 13.7.
- have applied iOS 14 promise fix.

Somnus007:
Same here. Any one can helps?

sidferreira:
Not hanging, but returns `[]`

garrettjoecox:
For us the purchase dialog shows, and the purchase listener gets called correctly, but the `requestSubscription` call returns a promise that is never resolved.

garrettjoecox:
Seems like this might have been resolved here https://github.com/dooboolab/react-native-iap/pull/1064

the-dut:
I applied that promise fix previously, so that's not the problem here.

compelling:
Here's how I did this: 
` // request purchase, listener registered above will receive notification when processing done
        RNIap.requestSubscription(this.productIds[0]).catch(err => {
            console.log(err.code, err.message);
        });`

the-dut:
I was calling the method with the boolean parameter false, which I think is the recommendation.  Either way I am still seeing very strange behavior on iOS14 no matter how I call the method, and it works fine on an iPad running iOS 13.7.  I tested requestSubscription with with and without the boolean and got two different behaviors, seemingly random.  But I am never getting the purchase dialogs.  Sometimes it never calls the RequestUpdatedListener.  Sometimes it calls it immediately with what looks like an invalid receipt.  

Are there others who are having no problems with their subscriptions?

theDev23:
Hey, do you any of you use `getPurchaseHistory()` or `getAvailablePurchases()` just before `requestSubscription()` if so, this seem to be the problem on iOS14.

the-dut:
[...]

PR #1227, opened 2021-01-07
Fix posible problem clearTransactionsIOS promise

andresordonezfm:
Solution recommended by rusty120 to avoid a possible deadlock in the promise of the clearTransactionsIOS

https://github.com/dooboolab/react-native-iap/issues/1120#issuecomment-754840685


hyochan (approved):
Wow thanks for bringing up the issue and find out the solution 🙌

PR #1228, opened 2021-01-09
Add sandbox agnostic receipt verification

ninjaboy:
As per [Apple receipt verification best practices](https://developer.apple.com/documentation/appstorereceipts/verifyreceipt): 
> As a best practice, always call the production URL for verifyReceipt first, and proceed to verify with the sandbox URL if you receive a 21007 status code.

In case if `isTest` parameter is `undefined` or defined as `null` (basically means that developer prefers to perform environment agnostic check:
1. Query production verification endpoint
2. In case if 20017 error is returned - query sandbox receipt verification endpoint

For the cases when `isTest` parameter is explicitly defined as `true` or `false` - run the old logic

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -649,7 +649,7 @@ export const finishTransaction = (
         } else if (
           purchase.userIdAmazon ||
           (!purchase.isAcknowledgedAndroid &&
-          purchase.purchaseStateAndroid === PurchaseStateAndroid.PURCHASED)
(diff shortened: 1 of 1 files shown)


hyochan:
@iaphub Could you kindly review this? Looks like this is somehow related to what you are working on.

hyochan (approved):
I've just gone through this and this is what Apple wanted. Thanks for your contribution 💯

Issue #351, opened 2018-12-28
what is itemSkus? package name or product id?


Nits7029:
### Version of react-native-iap
latest

### Version of react-native
0.56

### Platforms you faced the error (IOS or Android or both?)
Android

### Expected behavior
it should return products defined in play console

### Actual behavior
getProducts empty array

### Tested environment (Emulator? Real Device?)
Real Device

### Question
RNIap.getProducts(itemSkus) -> what to give in itemSkus? its package name or product id defined in in-app products?

here is my test products i have created and i have published an app so all products are active.

![screen shot 2018-12-28 at 11 13 56 am](https://user-images.githubusercontent.com/9457796/50504277-77079780-0a92-11e9-8f39-f15395c99626.png)

Nits7029:
got it, its product id. 

there are default product id's provided android.test.purchased, android.test.canceled, android.test.item_unavailable

so if you update your readme for the same then it would be great for other users to quickly integrate.

Thanks

PR #1239, opened 2021-01-29
Update README.md

jacquesdev:
Update references to `items` to refer to `products`, otherwise the term is too generic and for someone new to subscriptions this is not clear. Also related to https://github.com/dooboolab/react-native-iap/issues/351


hyochan (approved):
LGTM!

Issue #1203, opened 2020-12-07
getProducts() returns only Product fields: price, currency, productId


OlegPanfyorov:
### Version of react-native-iap
5.2.0

### Version of react-native
0.63.3

### Platforms you faced the error (IOS or Android or both?)
IOS

### Expected behavior
When I call getProducts() or getSubscriptions() it should return me a Product objects array with all fields described in docs similar to this

`{ 
    productId: 'com.XXX.sdk38.annual',
    subscriptionPeriodUnitIOS: 'YEAR',
    description: 'Activate Premium features on XXX',
    introductoryPrice: '',
    title: 'Annual',
    introductoryPriceSubscriptionPeriodIOS: '',
    introductoryPriceNumberOfPeriodsIOS: '',
    discounts: 
     [ { localizedPrice: '€0,00',
         numberOfPeriods: '1',
         type: 'SUBSCRIPTION',
         identifier: 'com.XXX.sdk38.tryannual',
         paymentMode: 'FREETRIAL',
         price: '0',
         subscriptionPeriod: 'MONTH' } ],
    type: 'Do not use this. It returned sub only before',
    localizedPrice: '€54,99',
    introductoryPricePaymentModeIOS: '',
    price: '54.99',
    currency: 'EUR',
    subscriptionPeriodNumberIOS: '1'
 }`

### Actual behavior
When I call getProducts() or getSubscriptions() it returns me a product with only 3 fields (see attached screenshot):
- price
- currency
- productId
` {price: "19.99", currency: "USD", productId: "hiii.premium1a"}`

### Tested environment (Emulator? Real Device?)
Real device. Production scheme with build configuration Debug and Debug executable set to true. But generally, we've noticed the same behavior on prod.

NSNJRGL:
Hi @OlegPanfyorov, can you get items on android using getProducts()? If yes, how you solve it? Thanks :)

TimPetricola:
I've had the same problem when upgrading from 5.1.1 to 5.1.3. It doesn't affect every device/account, but we saw it happening in production for ~800 users. I could not find a specific pattern, but reverting to 5.1.1 fixed it.

calluna18:
We have also noticed this issue for random users after 5.1.2. I noticed that it seems like it is the properties after the newly added `countryCode` that gets left out so probably an issue related to this.
https://github.com/dooboolab/react-native-iap/pull/1186/files#

[...]

PR #1245, opened 2021-02-06
Revert "Add iOS store country code"

ninjaboy:
Reverts dooboolab/react-native-iap#1186 due to https://github.com/dooboolab/react-native-iap/issues/1203
This seems to be a very optional field for now, but the change seems to be introducing issues in the production.
Let's revert this for now and release a new version please.


hyochan:
Could someone verify of this really solves the problem?

PR #1246, opened 2021-02-06
Prevent nil countryCode

jochem725:
This should prevent #1203 from happening.


hyochan (commented):
Looks fine to me. Let's see how it goes.

PR #1267, opened 2021-02-28
Fix: remove IAPPromotionObserver

imjihun:
- remove IAPPromotionObserver from xcode project
- it is for installing by manual


hyochan (approved):
LGTM!

PR #1277, opened 2021-03-16
Adding introductoryPriceAsAmountAndroid

Kelvne:
We already have the iOS version for this property, I'm sure in the future both should be renamed just for `introductoryPriceAsAmount`.

But to avoid breaking changes I think this works for now, my team is already using this fork in production because we needed this for both platforms.


hyochan (approved):
Acceptable 👍

PR #1302, opened 2021-04-15
[Amazon] Add promise handling for Amazon purchases instead of resolving immediately.

curiousdustin:
Without this addition, `buyItemByType` on Amazon immediately returns a resolved promise.

This change rejects or resolves the promise similar to how the Google Play implementation does.

### DIFF
--- a/android/src/main/java/com/dooboolab/RNIap/RNIapAmazonListener.java
+++ b/android/src/main/java/com/dooboolab/RNIap/RNIapAmazonListener.java
@@ -199,6 +199,14 @@ public class RNIapAmazonListener implements PurchasingListener {
         if (response.hasMore()) {
           PurchasingService.getPurchaseUpdates(false);
         } else {
+          if (purchases.size() > 0 && promiseItem != null) {
+            DoobooUtils
+              .getInstance()
+              .resolvePromisesForKey(
+                RNIapAmazonModule.PROMISE_BUY_ITEM,
+                promiseItem
+              );
+          }
           DoobooUtils
             .getInstance()
             .resolvePromisesForKey(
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/dooboolab/RNIap/RNIapAmazonListener.java
```diff
@@ -199,6 +199,14 @@ public void onPurchaseUpdatesResponse(
...
           PurchasingService.getPurchaseUpdates(false);
         } else {
+          if (purchases.size() > 0 && promiseItem != null) {
+            DoobooUtils
+              .getInstance()
+              .resolvePromisesForKey(
+                RNIapAmazonModule.PROMISE_BUY_ITEM,
+                promiseItem
+              );
+          }
```
hyochan, line 209: Don't you want to `return` here to prevent continuing code execution?

curiousdustin: No, in this case, we want to potentially resolve all three possible promises that triggered this `onPurchaseUpdatesResponse` response from the Amazon SDK. (there are 2 resolves right after this)

hyochan: Do you mean you want to resolve them at once? Only one event should resolve one promise at the same time.

curiousdustin: Maybe I am just naive, but why not? 

If 3 separate promises have been created, and are pending, and this single callback from the Amazon SDK contains the necessary data to resolve them, why not resolve all of them at that point?

Also, there are already 2 resolves in the code immediately after this, so if this really is an issue, those should be changed as well.

hyochan: @curiousdustin I think you've thought like below.

image

While I think is below.
image

When you call the `buyItemByType`, it opens the promise to be resolved. However, when other thread tries to resolve promise if not opened, I am afraid some miss leading behavior will occur.

Secondly, opening a promise to resolve then if 3 promises are ready to be resolved at a single operation, it doesn't look right to me.

hyochan: @curiousdustin I was going to update these changes myself. However, I am also not sure how this is going to cause the side-effect. Currently, I've changed my mind to accept this `PR`.

Let's see who's coming after this 🙏

hyochan (approved):
Thanks for the contribution 🙌

PR #1303, opened 2021-04-16
add force refresh receipt for ios 

sbin1211:
add force refresh receipt for ios to fix this kind of problems

An empty in_app array indicates that StoreKit has not recorded any transactions for that user yet. It may be that the application receipt has not yet been updated. When this happens, your app can inform the user that the receipt does not appear current and ask whether to refresh it

pls take a reference from 
https://stackoverflow.com/questions/36010595/itunes-reciept-validation-in-app-empty#:~:text=3-,An%20empty%20in_app%20array%20indicates%20that%20StoreKit%20has%20not%20recorded,ask%20whether%20to%20refresh%20it.
and 

https://developer.apple.com/library/archive/technotes/tn2413/_index.html#//apple_ref/doc/uid/DTS40016228-CH1-RECEIPT-MY_APP_VALIDATES_ITS_RECEIPT_WITH_THE_APP_STORE_VIA_PAYMENTQUEUE_UPDATEDTRANSACTIONS__AFTER_A_SUCCESSFUL_PURCHASE__HOWEVER__THE_RETURNED_RECEIPT_CONTAINS_AN_EMPTY_IN_APP_ARRAY_RATHER_THAN_THE_EXPECTED_PRODUCTS_

### DIFF
--- a/ios/RNIapIos.m
+++ b/ios/RNIapIos.m
@@ -298,9 +298,10 @@ RCT_EXPORT_METHOD(buyPromotedProduct:(RCTPromiseResolveBlock)resolve
     }
 }
 
-RCT_EXPORT_METHOD(requestReceipt:(RCTPromiseResolveBlock)resolve
+RCT_EXPORT_METHOD(requestReceipt:(BOOL)refresh
+                  resolve:(RCTPromiseResolveBlock)resolve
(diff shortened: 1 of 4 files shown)


hyochan (approved):
Thanks for the great research and the contribution :pray:

PR #1324, opened 2021-05-07
[Typescript] Strict type errors

andresesfm:
When enabling strict mode, several errors surfaced. This PR takes care of them by providing sensible default return alternatives

### DIFF
--- a/src/iap.ts
+++ b/src/iap.ts
@@ -81,11 +81,13 @@ function getAndroidModule(): any {
 function checkNativeAndroidAvailable(myRNIapModule: any): Promise<void> {
   if (!myRNIapModule)
     return Promise.reject(new Error(IAPErrorCode.E_IAP_NOT_AVAILABLE));
+  return Promise.resolve()
 }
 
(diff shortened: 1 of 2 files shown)


src/iap.ts
```diff
@@ -108,14 +110,14 @@ export const initConnection = (): Promise<boolean> =>
 
       return myRNIapModule.initConnection();
     },
-  })();
+  }) || Promise.resolve)();
```
hyochan, line 113: Don't you think we should reject it here instead of resolving it? Like below.
```
Promise.reject('Current platform is not supported in react-native-iap');
```

Also, same as other changes.

andresesfm: Hello @hyochan , thank you for looking at this PR. The reason I went with resolve and not reject was because it's consistent with the other responses in that method here: ```if (!RNIapIos) return Promise.resolve();``` and here: ```if (!RNIapModule || !RNIapAmazonModule) return Promise.resolve();```

hyochan (approved):
LGTM!

PR #1333, opened 2021-05-11
fix lint issues [Android]

andresesfm:
Includes deprecation warnings

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,11 +2,11 @@
 buildscript {
   repositories {
     google()
-    jcenter()
+    mavenCentral()
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:4.1.0'
+    classpath 'com.android.tools.build:gradle:4.2.0'
   }
 }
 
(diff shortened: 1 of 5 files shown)


android/build.gradle
```diff
@@ -2,11 +2,11 @@
 buildscript {
   repositories {
     google()
-    jcenter()
```
hyochan, line 5: Could you verify that this change is correct? Previously, this change made the module broken and fixed in #1328

android/build.gradle
```diff
@@ -41,8 +41,8 @@ android {
 }
 
 repositories {
-  mavenCentral()
   google()
+  mavenCentral()
```
hyochan, line 45: What is the actual effect of changing this order?

android/build.gradle
andresesfm: Hello @hyochan, I noticed that, thank you for fixing it last time. This time though, I have made sure everything works by re-running the new CI check for android https://github.com/dooboolab/react-native-iap/pull/1333/checks?check_run_id=2602685566 now we can move faster with more confidence

android/build.gradle
andresesfm: Just a bit better speed when building

hyochan (approved):
LGTM!

PR #1335, opened 2021-05-13
[Amazon] register listener on main thread

andresesfm:
Amazon SDK complains when registering the Purchase Listener on any thread other than main (UI)


hyochan (approved):
LGTM!

PR #1356, opened 2021-05-27
fire tv detection

andresesfm:
Aside from the installer, we can detect if the app has been installed on an Amazon device by following the guidelines described here: https://developer.amazon.com/docs/fire-tv/identify-amazon-fire-tv-devices.html
Especially useful when sideloading the app for testing purposes

Cleaner than the suggested fallback logic: 
```
if (__DEV__) {
  RNIap.setFallbackInstallSourceAndroid(RNIap.InstallSourceAndroid.AMAZON);
}
...
```

### DIFF
--- a/IapExample/package.json
+++ b/IapExample/package.json
@@ -13,7 +13,7 @@
     "apsl-react-native-button": "^3.1.1",
     "react": "16.9.0",
     "react-native": "0.61.5",
-    "react-native-iap": "^5.0.1",
+    "react-native-iap": "^6.0.6",
     "react-navigation": "^4.3.9"
   },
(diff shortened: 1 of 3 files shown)


hyochan (approved):
LGTM!

nochkin:
Just to add 2 cents here.
The fallback procedure is only when everything else failed and actually good for dev purposes mainly. For example when you sideloaded an app and want to force using Amazon or Google IAP procedures.

If your Amazon Fire TV has GAPPS installed and you want to force using those, you can use the fallback when your app is sideloaded. An automatic detection would not know what you are testing against.
Please keep this use-case in mind.

PR #1358, opened 2021-05-28
[Android] Remove buildToolsVersion

dulmandakh:
Setting buildToolsVersion is no longer recommended, it'll default to Android Gradle Plugin defaults. It makes DX worse by requiring multiple versions of build tools by various dependencies require various versions.


hyochan (approved):
LGTM!

PR #1377, opened 2021-06-15
Fix a wrong package.json setting

ifsnow:
The value of `main` property in `package.json` is "index.ts".
 I think this is problematic. For example, `Flow` throws a "Cannot resolve module" error.


hyochan (approved):
OMG! This is definitely a problem. It is also a wrong config on `ts` side.

Thank you for the fixes :pray:

PR #1379, opened 2021-06-15
First unit test

andresesfm:
- Added jest
- initConnection Google Play first test
- test on CI
- fix lint


hyochan (approved):
Thanks for coming back and fixing out :1st_place_medal: 

Just one question. Can we remove all comments in `jest.config.js` not being used?

PR #1380, opened 2021-06-16
fixes java lint errors and adds check to CI

andresesfm:


.github/workflows/ci.yml
```diff
@@ -28,14 +28,15 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 1.8
-      uses: actions/setup-java@v1
+    - name: Set up JDK 11
```
hyochan, line 31: Why 11 and not 14 if you want to upgrade?

hyochan (approved):
LGTM!

.github/workflows/ci.yml
andresesfm: 11 is the one that ships with Android studio and the one needed to run the Google check style gradle plugin 😄

Issue #1210, opened 2020-12-22
The operation couldn’t be completed. (SKErrorDomain error 2.)


AdamGerthel:
### Version of react-native-iap
4.6.3

### Version of react-native
0.63.2

### Platforms you faced the error (IOS or Android or both?)
iOS

### Expected behavior
Purchase succeeds

### Actual behavior
Error occurs, see https://sentry.io/share/issue/d8d09e708b064ae5a9639ab5bbb13cbc/

### Tested environment (Emulator? Real Device?)
Real device

### Steps to reproduce the behavior
I've only come across this in production logs (i.e. real users are affected), and I haven't been able to reproduce it. It started appearing in logs after I upgraded from 4.5.3 to 4.6.3.

chrisozgo99:
Facing same issue

AdamGerthel:
Question: Are we supposed to run `finishTransaction` on failures as well? I'm currently only using it in `purchaseUpdatedListener` and not in `purchaseErrorListener`, but [this](https://developer.apple.com/forums/thread/51621) thread seems to indicate that I should (although that's not RN-iap specific).

hyochan:
> Question: Are we supposed to run `finishTransaction` on failures as well? I'm currently only using it in `purchaseUpdatedListener` and not in `purchaseErrorListener`, but [this](https://developer.apple.com/forums/thread/51621) thread seems to indicate that I should (although that's not RN-iap specific).

[The link] best answers your question. Shortly, Your app needs to finish every transaction, regardless of whether the transaction succeeded or failed.

AdamGerthel:
@hyochan so the [example app](https://github.com/hyochan/DoobooIAP/blob/master/src/components/screen/Main.tsx) is doing it wrong as well? My own app implements RN-iap in the same way as the sample.

hyochan:
@AdamGerthel I don't think this is a problem with the `example` code. When does that error message appear? During the purchase?

PR #1385, opened 2021-06-22
Fix an unnecessary error when canceling a purchase

ifsnow:
Related to #1210

If I cancel during purchase interaction, the following error occurs.
```
Possible Unhandled Promise Rejection (id: 0):
Error: The operation couldn’t be completed. (SKErrorDomain error 2.)
promiseMethodWrapper@http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:2220:45
http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:201046:43
step@http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:200702:25
http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:200632:20
http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:200604:73
tryCallTwo@http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:3446:9
doResolve@http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:3610:25
Promise@http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:3469:14
http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:200583:36
ios@http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:201041:25
requestSubscription@http://192.168.0.2:8081/index.bundle?platform=ios&dev=true&minify=false:201058:27
```

I think It'd be better to ignore errors caused by user cancellation.

### DIFF
--- a/ios/RNIapIos.m
+++ b/ios/RNIapIos.m
@@ -461,9 +461,12 @@ RCT_EXPORT_METHOD(presentCodeRedemptionSheet:(RCTPromiseResolveBlock)resolve
                                              ];
                         [self sendEventWithName:@"purchase-error" body:err];
                     }
-                    [self rejectPromisesForKey:transaction.payment.productIdentifier code:[self standardErrorCode:(int)transaction.error.code]
-                                       message:transaction.error.localizedDescription
-                                         error:transaction.error];
+
+                    if (transaction.error.code != SKErrorPaymentCancelled) {
+                        [self rejectPromisesForKey:transaction.payment.productIdentifier code:[self standardErrorCode:(int)transaction.error.code]
+                                        message:transaction.error.localizedDescription
+                                            error:transaction.error];
+                    }
                 });
                 break;
             }


hyochan (approved):
Reasonable! Thanks for the fixes :)

yuri-lomashko-itechart:
Actually that PR made breaking change, because it changed the default logic of purchase promise with `Cancel` error. It's critical for applications which respect not only `purchaseUpdatedListener` and `purchaseErrorListener` listeners but also promise of purchase process.

Also as I understand it's a memory leak, because when user cancel the IAP modal the promise will be in PENDING status forever.

Besides, I don't see any reasons why the error should be ignored on promise side because this error is a part of IAP api in fact
@hyochan Can we rollback these changes?

ifsnow:
@hyochan If there are existing users who are having problems with this breaking change, I think you should consider a rollback.

yuri-lomashko-itechart:
> Can you please rollback these changes?

resolved in #1504

Issue #555, opened 2019-06-22
Android: Crashes when providing Old Subscription ID and Proration Mode = DEFERRED


mifi:
### Version of react-native-iap
3.0.0

### Version of react-native
0.59.2

### Platforms you faced the error (IOS or Android or both?)
Android

### Expected behavior
App should not crash

### Actual behavior
App crashes after completing android purchase dialog

### Tested environment (Emulator? Real Device?)
Huawei P Smart 2018

### Steps to reproduce the behavior
- Create 2 subscriptions in Google Play, one monthly and one yearly
- Subscribe to monthly subscription. Works perfectly
- Subscribe to yearly subscription, with `Old Subscription ID` and `Proration Mode = DEFERRED` (4) sent to `RNIap.requestSubscription`
- App crashes without any java stack trace in react-native log-android
- NOTE! It works perfectly when not specifying any Proration Mode, so it is related to DEFERRED

By the way awesome job on this module! I'm happy that someone abstracted these crazy complex APIs a bit. This issue is not a dealbreaker because I can use other Proration Mode than DEFERRED but I thought I'd report it, and maybe someone can understand why this happens.

hyochan:
Thanks for the detailed issue. Let's wait for those who can help us.

iaphub:
See [this comment](https://github.com/dooboolab/react-native-iap/issues/888#issuecomment-660519819) regarding your issue with deferred subscriptions.

andresesfm:
Documentation: https://github.com/dooboolab/react-native-iap/pull/1387

Issue #888, opened 2019-12-28
Error when upgrade/downgrade subscription on Android with DEFERRED mode


howg0924:
### Version of react-native-iap

4.3.3

### Version of react-native

0.59.10

### Platforms you faced the error (IOS or Android or both?)

Android

### Expected behavior

When upgrade/downgrade a subscription with DEFERRED mode, purchase update listener should be called with new transaction receipt.

### Actual behavior

purchase error listener is called with following error:

{ message: 'purchases are null.',
code: 'OK',
debugMessage: '',
responseCode: 0 }

### Tested environment (Emulator? Real Device?)

Real Device - sandbox

### Steps to reproduce the behavior

1.Purchase a subscription with RNIap.requestSubscription(sku), this works correctly.

2.Upgrade or downgrade the subscription with RNIap.requestSubscription(newSku, false, sku, 4). (4 is the DEFERRED Proration Mode) Google Play billing dialog will show the subscription plan is changed successfully, and you will receive a notification email from Google Play about that. However, the error listener is called with the above mentioned error.

3.If you call RNIap.requestSubscription(newSku, false, sku, 4) again, Google Play billing dialog will say they cannot change the subscription plan, and the error listener is called with the following error:

{ message: 'Google is indicating that we have some issue connecting to payment.',
  code: 'E_DEVELOPER_ERROR',
  debugMessage: '',
  responseCode: 5 }

But sometimes Google Play billing dialog will say your order is under processing and your product should be delivered soon, and the error listener is called with the following error:

{ message: 'You already own this item.',
  code: 'E_ALREADY_OWNED',
  debugMessage: '',
  responseCode: 7 }

I guess this is due to the transaction of step 2 is waiting to be acknowledged. However, since step 2 does not return a receipt, we cannot acknowledge it.

hyochan:
Try to call `getAvailablePurchases` after you've upgrade/downgrade subscription.
Refer to [doc](https://developer.android.com/google/play/billing/billing_subscriptions#Allow-upgrade).

howg0924:
I tried getAvailablePurchases() after upgrade/downgrade the subscription, but unfortunately, only the transaction receipt of original purchase is in the returned array.

RNIap v2.3.19 & v2.5.5 works correctly, buySubscription() will return a new receipt when upgrade/downgrade the subscription. But I would like to upgrade to v3 or v4 which has a better event based model if possible.

[...]

PR #1387, opened 2021-06-22
Adds documentation to Android deferred proration

andresesfm:
Credit goes to @Paduado and @hyochan 
See discussion here: https://github.com/dooboolab/react-native-iap/issues/888
It adds documentation so now we can close: 
https://github.com/dooboolab/react-native-iap/issues/555

### DIFF
--- a/android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
+++ b/android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
@@ -65,7 +65,7 @@ public class RNIapModule extends ReactContextBaseJavaModule implements Purchases
   public RNIapModule(ReactApplicationContext reactContext) {
     super(reactContext);
     this.reactContext = reactContext;
-    this.skus = new ArrayList<SkuDetails>();
+    this.skus = new ArrayList<>();
     reactContext.addLifecycleEventListener(lifecycleEventListener);
   }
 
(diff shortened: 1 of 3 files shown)


android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
```diff
@@ -199,21 +199,13 @@ private void consumeItems(
...
         new Runnable() {
           @Override
           public void run() {
-            final WritableNativeArray array = new WritableNativeArray();
             Purchase.PurchasesResult result =
                 billingClient.queryPurchases(BillingClient.SkuType.INAPP);
-            if (result == null) {
-              promise.reject("refreshItem", "No results for query");
-              return;
-            }
```
hyochan, line 216: Won't we have any downside removing this? If then, can we also remove 
```
Purchase.PurchasesResult result = billingClient.queryPurchases(BillingClient.SkuType.INAPP);
```

android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
```diff
@@ -232,14 +224,8 @@ public void flushFailedPurchasesCachedAsPending(final Promise promise) {
...
           @Override
           public void run() {
-            final WritableNativeArray array = new WritableNativeArray();
             Purchase.PurchasesResult result =
                 billingClient.queryPurchases(BillingClient.SkuType.INAPP);
-            if (result == null) {
-              // No results for query
-              promise.resolve(false);
-              return;
-            }
```
hyochan, line 242: Same here. How about putting `@NonNull` decorator?

hyochan (commented):
Could you kindly check above comments?

android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
andresesfm: Hello @hyochan, no it won't have any effect at all since `array` is never used and `result` is non-nullable so that check is also not needed. We do however use result so we cannot remove it's assignment

android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
andresesfm: Same as in the other one, `array` is created but never used and `result==null` is never true as `queryPurchase` return is non-nullable

hyochan: Thanks for the reply!! Understood 👍

hyochan (approved):
Thanks for your contribution as usual :)

PR #1388, opened 2021-06-22
Fixes race condition caused by billing client dereference

andresesfm:
There are many tickets related to duplicate events and null pointer exceptions. They were caused by the implementation that constantly created a new listener. In this PR I:
1) Standardize Android calls by checking for connection to be initialized before calling any of the other methods
2) Check for duplicate calls to `initConnection` that could cause a race condition
3) Add documentation to the proper way to initialize/tear down the connection


andresesfm:
This needs more work. Reading the documentation for the iap library v3, I realized that we should guard the existence of the client instead of creating a new one when needed. This was likely implemented as such catering to older versions of t he SDK

hyochan:
@andresesfm Is this ready for review?

lc3t35:
Hi, thank you @andresesfm nice catch ! I'm stuck on Android for all the tests related to purchasing (ios pass the tests (highfive) ) ;) @hyochan Hope you can review and validate the PR as soon as possible. Thank you for your time !

hyochan (approved):
Updated small fixes in docs.

Thanks for the contribution again :raised_hands:

Issue #958, opened 2020-03-24
Navigating from react-native-app to playstore for cancelling in-app Subscriptions using particular url doesn't work


LohithaYalavarthi:
### Version of react-native-iap
2.4.8

### Version of react-native
0.59.10

### Platforms you faced the error (IOS or Android or both?)

Android

### Expected behavior
When  using this url to navigate to playstore subscription page from react native app it should be able to navigate it
        'https://play.google.com/store/account/subscriptions?package='dummypackage'&sku=box100.subscription.dev' and using the linking api of react-native 

Instead It can only  navigate using 'https://play.google.com/store/apps/details?id=packagename'

### Actual behavior
'https://play.google.com/store/account/subscriptions?package='dummypackage'&sku=box100.subscription.dev' and using the linking api of react-native

this url doesnt work

### Tested environment (Emulator? Real Device?)
Android Oppo 5s

### Steps to reproduce the behavior
Use the code below
navigateToPlaystore = productId => {
    console.log('getDevicePlatform() :', getDevicePlatform(), productId)
    let url = ''
    if (getDevicePlatform() === 'android') {
      url =
        'https://play.google.com/store/account/subscriptions?package=com.peerzacademy.peertutor.dev&sku=box100.subscription.dev'
        } else if (getDevicePlatform() === 'ios' && Platform.version >= 12) {
      url = 'https://apps.apple.com/account/subscriptions'
    } else {
      url =
        'https://buy.itunes.apple.com/WebObjects/MZFinance.woa/wa/manageSubscriptions'
    }
    Linking.canOpenURL(url)
      .then(supported => {
        if (!supported) {
          console.log("Can't handle url: " + url)
        } else {
          return Linking.openURL(url)
        }
      })
      .catch(err => console.error('An error occurred', err))
  }

erickgtzh:
were you able to navigate to your in-app Subscriptions from react native??

Issue #725, opened 2019-09-21
Possible issue with requestPurchase promise


doomsower:
This is not a confirmed bug, since I'm rewriting my app to v3/v4 and haven't tested it manually yet. These are just some observations and a feature request.

Firstly, I've examined iOS and Android `requestPurchase`promises and found inconsistency. On iOS, promises are resolved on [per-purchase basis](https://github.com/dooboolab/react-native-iap/blob/67d3720024b6c0464dd47550562e3da2b26b67b0/ios/RNIapIos.m#L445): 

```objc
[self resolvePromisesForKey:RCTKeyForInstance(transaction.payment.productIdentifier) value:purchase];
```

On Android it seems that there's [one promise](https://github.com/dooboolab/react-native-iap/blob/9bcc75ce2774644b32fdcd6d845e64e0a223741d/android/src/main/java/com/dooboolab/RNIap/RNIapModule.java#L516) for all `requestPurchase` calls:

```java
DoobooUtils.getInstance().resolvePromisesForKey(PROMISE_BUY_ITEM, item);
```

I'm not deeply familiar with IAP flows, but it seems that if user tries to buy 2 products in quick succession, then one of the `requestPurchase` calls might never reject or resolve. Or maybe it'll throw because of resolving same promise twice. Why not use product identifier as promise key as on iOS?

Secondly, a related issue. When `requestPurchase` rejects, it's possible to know which purchase has failed, because `requestPurchase` arguments are in the same scope. However, `purchaseErrorListener` receives only the error code and message, so there's no way to know which purchase failed. So I think `productId` field should be added to `PurchaseError`.

### Version of react-native-iap

`next` branch

AustP:
I am using version 3.4.12 and am having an issue with `requestPurchase`'s promise on Android.

The problem I am having with it is that it never resolves after going through the purchase process. Interestingly, it rejects if I cancel the process halfway, but if I finish the process, it never resolves.

hyochan:
I've reviewed #724. `requestPurchase` should not resolve `promises` in the next major release which you can use by `yarn install react-native-iap@next`.

doomsower:
What about the second part of this issue?

> However, purchaseErrorListener receives only the error code and message, so there's no way to know which purchase failed. So I think productId field should be added to PurchaseError.

Without this `purchaseErrorListener` is useless and I have to use promises.

hyochan:
> What about the second part of this issue?
> 
[...]

PR #1389, opened 2021-06-22
Remove an unnecessary reject when canceling by user

ifsnow:
Related to #1385

I should have worked together in my previous PR, but it's missing.

Unlike iOS, Android does not throw warning errors, but It'd be better to be excluded for the same reason.

I don't think it's necessary to release a new version right now. It's enough to be included in the next version.

### DIFF
--- a/android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
+++ b/android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
@@ -518,7 +518,8 @@ public class RNIapModule extends ReactContextBaseJavaModule implements Purchases
                   BillingFlowParams.ProrationMode.IMMEDIATE_AND_CHARGE_PRORATED_PRICE);
               if (!type.equals(BillingClient.SkuType.SUBS)) {
                 String debugMessage =
-                    "IMMEDIATE_AND_CHARGE_PRORATED_PRICE for proration mode only works in subscription purchase.";
+                    "IMMEDIATE_AND_CHARGE_PRORATED_PRICE for proration mode only works in"
+                        + " subscription purchase.";
                 WritableMap error = Arguments.createMap();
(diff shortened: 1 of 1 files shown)


andresesfm:
@ifsnow The CI build fails at the lint check for formatting. To fix please run `./gradlew goJF` in the android folder 🙏

ifsnow:
@andresesfm Thanks for your feedback!

I used [google-java-format](https://formulae.brew.sh/formula/google-java-format) instead of `./gradlew goJF` as a jvm version issue.

andresesfm:
@ifsnow Thank you for formating your PR. I have a question for you: Since we are not rejecting the Promise anymore, would it make sense to resolve it with a value? 
@hyochan you are more familiar with the code base but I think this would leave the promise hanging in memory, don't you think?

ifsnow:
@andresesfm In my opinion, it could be logic to maintain backwards compatibility. If we only use `purchaseErrorListener` for error handling, we probably don't need the promise logics.

The opinion of @hyochan who knows the history well seems to be important.

hyochan (commented):
Thanks for the discussion. Like @ifsnow pointed out, it matters with backward compatibility and even correct that `purchaseErrorListener` should handle them. I was stealthily deprecating the `promises` since this will not cover the right lifecycle as introduced in [version 3](https://medium.com/dooboolab/react-native-iap-v3-1259e0b0c017). 

However, some people still needed them for some reason (mostly for checking loading status) in issues like #1058 #958 #725.

I still don't recommend implementation with `promises`. However, I just supported them just for exposing the possibilities. For this reason, I am not really opinionated with these kinds of discussions but IMO, I feel it'd be better to resolve or reject promises in every case just to reduce the confusion.

hyochan:
If there is no further progress, I'll consider merging it soon!

hyochan (approved):

PR #1390, opened 2021-06-23
fix promise resolving logic

andresesfm:
Since the code split between stores, I created a separate file to deal with Google Play-specific logic for handling errors. However, this meant that I inadvertently created a separate copy of the hashmap that holds the promises. This could create a disconnect making some promises not resolve. This PR reverts that change by only using DoobooUtils as the host of such Promises store.


hyochan (approved):
Sounds like you are right! I've also couldn't grab this hole in the review :disappointed_relieved:

PR #1393, opened 2021-06-24
fixed regression on return type

andresesfm:


hyochan (approved):
Thanks for the fixes :+1:

PR #1401, opened 2021-07-02
Removed setInstallSource and updated docs

andresesfm:


hyochan (approved):
Thank you for the further supports on this.

PR #1407, opened 2021-07-06
Revert strict check for Amazon module

andresesfm:


hyochan (approved):
Thanks so much for the backup!

Issue #1409, opened 2021-07-08
After upgrading to v7.0.1 from 6.2.2 getPurchaseHistory does not return response


SergiOnGit:
> Please use [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want get some help out of it. Please report issue if there is a bug.

### Version of react-native-iap
7.0.1

### Version of react-native
0.64.2

### Platforms you faced the error (IOS or Android or both?)
Android. Not checked on IOS

### Expected behavior
It should return purchase history array

### Actual behavior
It doesn't return anything

### Tested environment (Emulator? Real Device?)
Both

### Steps to reproduce the behavior
Just upgrade to this version and call getPurchaseHistory()

What am i missing? It is working fine in 6.2.2.

Montezi:
Hi! I have the same issue  with almost same enviroment , with react native 0.63.4.

nikmedoed:
The same problem. `getAvailablePurchases()` also doesn't work.
`RN 0.59.8`, android

Workin on 6.3.0

Rolling back again

marcpechaitis:
Same here (on android device, works fine on iOS):
```
    const purchaseHistory = await RNIap.getPurchaseHistory();
    console.log('purchaseHistory', purchaseHistory);
    purchaseHistory.forEach((purchase) => {
      console.log('🔷🔷🔷🔷🔷 history', purchase.productId);
    });
 ```

`await RNIap.getPurchaseHistory();` never resolves/rejects, and the `console.log('purchaseHistory', purchaseHistory);` does not get called in 7.0.1

It works great in 6.3.0:
```
purchaseHistory [{"dataAndroid": <redacted>}]
[Fri Jul 09 2021 08:34:17.580]  LOG      🔷🔷🔷🔷🔷 history bingo.consumable.one
[Fri Jul 09 2021 08:34:17.670]  LOG      🔷🔷🔷🔷🔷 history bingo.consumable.two
[Fri Jul 09 2021 08:34:17.700]  LOG      🔷🔷🔷🔷🔷 history bingo.consumable.three
[Fri Jul 09 2021 08:34:17.700]  LOG      🔷🔷🔷🔷🔷 history bingo.nonconsumable.four
```

andresesfm:
I'm assuming you are all calling `initConnection()` beforehand, correct?. 
Also can somebody please post the adb logcat? I'm happy to take a look

marcpechaitis:
[...]

PR #1411, opened 2021-07-09
Fix getPurchaseHistory for Android (#1409)

marcpechaitis:


android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
```diff
@@ -399,7 +399,7 @@ public void onPurchaseHistoryResponse(
                   for (int i = 0; i < purchaseHistoryRecordList.size(); i++) {
                     WritableMap item = Arguments.createMap();
                     PurchaseHistoryRecord purchase = purchaseHistoryRecordList.get(i);
-                    item.putString("productId", purchase.getSkus().get(i));
+                    item.putString("productId", purchase.getSkus().get(0));
```
hyochan, line 402: Could you kindly give more information on how could this solve the problem?
Wouldn't these changes only return the first `productId`?

hyochan (changes requested):
@marcpechaitis Hope you to take a look at the review :pray:

android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
marcpechaitis: @hyochan sure.  In the [Google Play Billing Library v 4.0 Release Notes](https://developer.android.com/google/play/billing/release-notes), we learn that Google: 

> Added Purchase#getSkus() and PurchaseHistoryRecord#getSkus(). These replace Purchase#getSku and PurchaseHistoryRecord#getSku which have been removed.

`getSkus()` returns an array of skus, but we don't really have any information _why_ they made this change?  🤷‍♂️

> getSkus
> public ArrayList<String> getSkus ()
> Returns the product Ids.

In my app, I have 4 skus available for purchase, and my test user has purchased each one, but each one is returned as an item in `purchaseHistoryRecordList`, and the `purchase.getSkus()` call return an array with a single item.

The issue with the code using `.get(i)` on the results (`purchase.getSkus().get(i))`) is that once we start looping through the `purchaseHistoryRecordList` and `i` gets incremented, we are trying to `get()` items that _do not exist_.  When this occurs, the method fails and does not throw an error.  This is why the call to `getPurchaseHistory` in version 7.0.x does nothing.  I don't even seen anything in the logcat output when this happens (which is why I can't provide the output, as requested 😉 )

When we use the `.get(0)` on the results of `purchase.getSkus()` then we get results that are _consistent with the previous version of the library_, because the previous version of the library returned a single sku.

I suppose there are other things that we should consider?

1) We could try a nested loop on the results of `getSkus()` inside each item of the `purchaseHistoryRecordList` ?
 ```
for (int i = 0; i < purchaseHistoryRecordList.size(); i++) {
	WritableMap item = Arguments.createMap();
    PurchaseHistoryRecord purchase = purchaseHistoryRecordList.get(i);
    for (int j = 0; j < purchase.getSkus().size(); j++) {
		item.putString("productId", purchase.getSkus().get(j));
		item.putDouble("transactionDate", purchase.getPurchaseTime());
		item.putString("transactionReceipt", purchase.getOriginalJson());
		item.putString("purchaseToken", purchase.getPurchaseToken());
		item.putString("dataAndroid", purchase.getOriginalJson());
		item.putString("signatureAndroid", purchase.getSignature());
		item.putString("developerPayload", purchase.getDeveloperPayload());
		items.pushMap(item);
	}	
}
```
2) Can we wrap this logic in a `try()... catch()` to handle the error better?
3) How can we minimize disruption to developers who are expecting results the same as the earlier versions of the library... and also keep the payload the same as what iOS version returns?  Or do we create a new `getPurchaseHistoryAndroid()` method?

The bottomline, for me, is that when I make the simple change from `.get(i)` to `.get(0)` everything works again, just like it did in 6.3.0... but I'm happy to consider your feedback and help with the best solution for the community.

hyochan (approved):
Thanks for the detailed feedback. It is very clear now.

This is bad though that the length of `skus` is not identical to `purchaseHistoryRecordList`.

Your proposal looks great to me and that can be done in next `PR` I think.

I don't think you should worry too much about the breaking, we can add an additional readme to solve the problem for new version.

Issue #1424, opened 2021-07-14
Amazon App Store - App crashes on startup


JakobG-dev:
> Please use [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want get some help out of it. Please report issue if there is a bug.

### Version of react-native-iap
7.0.3

### Version of react-native
0.64.2

### Platforms you faced the error (IOS or Android or both?)
Android

### Problem
I am trying to add amazon IAP to my app.
The Amazon release version keeps crashing immediately after app starts. Debug version works without problems. The other release versions i have work without problem too.

### Tested environment (Emulator? Real Device?)
Tested on Android 10 Emulator and android 8 real device

### Steps to reproduce the behavior
This is the project: https://github.com/addiction-free-video/addiction-free-video

Attached find complete log from when starting app.
[adb_logcat.log](https://github.com/dooboolab/react-native-iap/files/6814633/adb_logcat.log)

andresesfm:
Thank you for the report @JakobG-dev I'll take a look

andresesfm:
@JakobG-dev this is my fix for the issue: https://github.com/dooboolab/react-native-iap/pull/1425 Could you please help us validate it?

JakobG-dev:
@andresesfm This didn't fix my issue, it just postponed the app from crashing until i call initConnection.

andresesfm:
Hello @JakobG-dev I looked at the SDK code and this https://github.com/dooboolab/react-native-iap/pull/1427 could help since the registration doesn't need to happen on the Main thread anymore. However, that still doesn't explain why it only happens to you on release. I looked at your project and it looks like you have included the proguard statements for the SDK. So can you please give this one a shot and report with a fresh adblog capture?. Please capture from the moment you launch the app as we should see
```
In-App Purchasing SDK initializing. SDK Version 2.0.76.4, IS_SANDBOX_MODE: ....
...
PurchasingListener registered: ...
PurchasingListener Context: ...
...
printed in them

JakobG-dev:
Hi @andresesfm, I tried #1427 and everything seems to work, it doesn't crash anymore.
Thank you so much! :pray:

[adb_logcat_working.txt](https://github.com/dooboolab/react-native-iap/files/6826164/adb_logcat_working.txt)

PR #1427, opened 2021-07-15
Remove init on main thread as it is no longer needed in this version of the SDK

andresesfm:
Fixes https://github.com/dooboolab/react-native-iap/issues/1424


hyochan (approved):
LGTM!

PR #1438, opened 2021-07-28
fixes subscription upgrade/downgrade

andresesfm:
androidOldSku is no longer required for subscription changes. The parameter was being mixed up with the purchase token. I attempted to deprecate it but since it was a required field before. Many clients would have to set it to null. The cleanest way to deal with it is to mark it as a breaking change and move on


hyochan (approved):
Thanks for being a saviour once again!

android/src/amazon/java/com/dooboolab/RNIap/RNIapAmazonModule.java
```diff
@@ -87,7 +87,6 @@ public void getPurchaseHistoryByType(final String type, final Promise promise) {
   public void buyItemByType(
       final String type,
       final String sku,
-      final String oldSku,
```
mars-lan, line 90: Dropping this without updating the TypeScript call site (https://github.com/dooboolab/react-native-iap/blob/master/src/iap.ts#L320) leads to an immediate crash.

PR #1447, opened 2021-08-05
Fix crash when init on Android Emulator

andresesfm:
When calling initConnection on an Android Emulator fails if not wrapped in catch. 

Resolving to false and returning prevents the crash and is more consistent with usage


hyochan (approved):
LGTM!!

PR #1456, opened 2021-08-07
Fix android crash by delaying ios check

### LINKED ISSUE #1454: Crash on android when calling getProducts()
"react-native-iap": "^7.2.0",
### Version of react-native
"react-native": "0.63.3",
Android
### Expected behavior
The function `getProducts()` should use the android module and return a list of products.
### Actual behavior
When calling `getProducts()` the function `getIosModule()` is called even if the OS is android. I then get the error `IAPErrorCode.E_IAP_NOT_AVAILABLE` throwned in `checkNativeiOSAvailable()`.
Real Device, Samsung SM-G950U1
1. Purchase a product on android
2. The error E_IAP_NOT_AVAILABLE is thrown

I'll try to push a fix ASAP!

andresesfm:
I went too far removing code, when not wrapping the `ios:` block with an async block was causing it to evaluate every time regardless of platform. The fix is to add the mentioned wraps that were previously removed

Closes #1454

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -22,10 +22,5 @@
   },
   "javascript.preferences.importModuleSpecifier": "relative",
   "typescript.preferences.importModuleSpecifier": "relative",
-  "prettier.configPath": ".prettierrc.js",
-  "workbench.colorCustomizations": {
-    "activityBar.background": "#441660",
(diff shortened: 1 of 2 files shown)


hyochan (approved):
Thanks for the quick fixes!

radko93:
@hyochan please release ASAP if possible

PR #1460, opened 2021-08-10
Added isCanceled to amazon properties

andresesfm:
Exposed `isCanceled` coming from Amazon's sdk. As there might be more than one receipt coming from purchase history


andresesfm:
Closing in favor of the more generic: https://github.com/dooboolab/react-native-iap/pull/1461

PR #1461, opened 2021-08-10
improve receipt types

andresesfm:
- Added isCanceled to amazon properties
- fix lint
- cleaned up types and refactored common code across platforms
- refactor receipt conversion for Amazon
- verbiage

### DIFF
--- a/IapExample/package.json
+++ b/IapExample/package.json
@@ -13,7 +13,7 @@
     "apsl-react-native-button": "3.1.1",
     "react": "17.0.1",
     "react-native": "0.64.2",
-    "react-native-iap": "file:..",
+    "react-native-iap": "7.2.1",
     "react-navigation": "4.3.9"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


IapExample/package.json
```diff
@@ -13,7 +13,7 @@
     "apsl-react-native-button": "3.1.1",
     "react": "17.0.1",
     "react-native": "0.64.2",
-    "react-native-iap": "file:..",
```
hyochan, line 16: Is there any reason why reverting this?

andresesfm: Yes, I've been testing and it turns out that Metro doesn't pick up on this notation. Reference here: https://github.com/facebook/metro/issues/68 I really wanted it to work, I ended up using wml: https://github.com/wix/wml#readme which is not great but it gets the job done. I really hope they enable linking on Metro at some point. In the meantime, there's no point in pointing to a file

hyochan: Thanks for spotting the issue. Have you also tried [this](https://github.com/facebook/metro/issues/68#issuecomment-876156490)?

Anyway, let's merge this `PR` and we may of this afterward.

hyochan (approved):
LGTM

PR #1465, opened 2021-08-13
migrated IapExample iOS to rn 0.64

andresesfm:


IapExample/App.js
```diff
@@ -116,7 +116,11 @@ class Page extends Component {
   async componentDidMount(): void {
     try {
       await RNIap.initConnection();
-      await RNIap.flushFailedPurchasesCachedAsPendingAndroid();
+      if(Platform.OS == 'android'){
+        await RNIap.flushFailedPurchasesCachedAsPendingAndroid();
+      }else{
+        await RNIap.clearTransactionIOS();
+      }
```
hyochan, line 123: Looks like our `linting` is not correctly configured here 🤔
```
if (Platform.OS === 'android') {
```

and 
```
} else {
```

would be the correct format.
Also is there any reason for using equality operator `==` and not identity operator `===` here?

andresesfm: Good catch @hyochan . I fixed the lint and the comparison. We are not linting js files, remember? you changed eslint to only check ts files a few weeks back. No big deal since we don't touch the example app that much but may be worth while to bring it back

hyochan: Thanks for the fixes. Also, I am not aware of the `example` project anymore. 
I am currently testing the latest `react-native-iap` in my [own repo here](https://github.com/hyochan/dooboolab.com/blob/master/src/components/pages/Sponsor.tsx) 😅

hyochan (approved):
LGTM!

PR #1476, opened 2021-09-01
Add quantity in purchase data

jonathangilbert02:
On iOS, we can `requestPurchaseWithQuantityIOS` which purchase a productId with a specific quantity.

However, in the purchase info when a purchase is completed, we only receive the productId.

I have done some changes so we can now receive both the productId and the requested quantity when a purchase is completed.

The quantity is fetched from the associated payment in the transaction. 

Documentation : https://developer.apple.com/documentation/storekit/skpayment/1506077-quantity?language=objc


hyochan (commented):
Thank you for the fixes!! Do you think we should add types under https://github.com/dooboolab/react-native-iap/blob/65803ec339f4146e011aeea4a7a230645dbdfb6d/src/types/index.ts#L58?

I think this can be `quantityIOS`.

jonathangilbert02:
Yes I will rename quantity to quantityIOS and add it under the types as suggested !

hyochan (approved):
LGTM!

PR #1482, opened 2021-09-03
Use React's Context to manage IAP state

r0b0t3d:
currently, every time calling `useIAP` hooks, it will call `initConnection` and register listeners with should only be done once when the app starts (as mentioning in the documentation). So I introduce new HOC `withIAPContext`

1. `withIAPContext` HOC responsible for
- Init context
- Init IAP connection
- Register listeners

2. `useIAP` same as before but removing the code that inits IAP and registers listeners.

## How to use
1. You have to wrap your app with the `withIAPContext` HOC

```jsx
import {withIAPContext} from 'react-native-iap';

const App = () => <View />;

export default withIAPContext(App);
```

2. Later then, somewhere in your components

```jsx
import {withIAP} from 'react-native-iap';

const YourComponent = () => {
  const {
    connected,
    products,
    promotedProductsIOS,
    subscriptions,
    purchaseHistories,
    availablePurchases,
    currentPurchase,
    currentPurchaseError,
    finishTransaction,
    getProducts,
    getSubscriptions,
    getAvailablePurchases,
    getPurchaseHistories,
  } = useIAP();

  return <View />;
};
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "react-native-iap",
   "version": "7.3.0",
   "description": "React Native In App Purchase Module.",
-  "main": "index.js",
+  "main": "index",
   "types": "index.d.ts",
   "postinstall": "dooboolab-welcome postinstall",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "react-native-iap",
   "version": "7.3.0",
   "description": "React Native In App Purchase Module.",
-  "main": "index",
+  "main": "index.js",
   "types": "index.d.ts",
   "postinstall": "dooboolab-welcome postinstall",
   "scripts": {
--- a/README.md
+++ b/README.md
@@ -101,6 +101,10 @@ Quick News
 Breaking Changes
 ----------------
 
+[8.0.0]
+- Now using React's Context to manage IAP state
+- Introduce `withIAPContext` HOC ([how to use](docs/docs/usage_instructions/using_hooks.md))
+
 [7.1.0]
 - `androidOldSku` is no longer required [#1438](https://github.com/dooboolab/react-native-iap/pull/1438).
 
--- a/docs/docs/usage_instructions/using_hooks.md
+++ b/docs/docs/usage_instructions/using_hooks.md
@@ -4,6 +4,45 @@ sidebar_position: 6
 
 # Using hooks
 
-From `react-native-iap@6.0.0+` we support the useIAP hook that handles purchases better. 
+## How to use hooks
+1. You have to wrap your app with the `withIAPContext` HOC
+
+```jsx
+import {withIAPContext} from 'react-native-iap';
+
+const App = () => <View />;
+
+export default withIAPContext(App);
+```
+
+2. Later then, somewhere in your components
+
+```jsx
+import {withIAP} from 'react-native-iap';
+
+const YourComponent = () => {
+  const {
+    connected,
+    products,
+    promotedProductsIOS,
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "react-native-iap",
   "version": "7.3.0",
   "description": "React Native In App Purchase Module.",
-  "main": "index.js",
+  "main": "index",
```
hyochan, line 5: Can you tell exact effect of changing this?

hyochan (commented):
Thanks for your interests and contribution!

However, I am not sure why we need this. Looks like everything in handled in `useIAP`.
We can use this without a context already like https://github.com/hyochan/dooboolab.com/blob/1c88b905a595bed12f7c80a3c6c02527d7c3a498/src/components/pages/Sponsor.tsx#L182-L190.

Could you kindly tell what makes context better here?

r0b0t3d:
Hi @hyochan, with `Context`, all below values are stored in `Context` and will be shared between components.
```
    connected,
    products,
    promotedProductsIOS,
    subscriptions,
    purchaseHistories,
    availablePurchases,
    currentPurchase,
    currentPurchaseError,
```

Let say I use `useIAP` in 2 different places
- In first component, I call `getSubscriptions`
```jsx
const Component1 = ( ) => {
  const { connected, subscriptions, getSubscriptions } = useIAP();

  useEffect(() => {
    if (connected) {
      getSubscriptions(skus);
    }
  }, [connected])
}
```
- In the second component, I would like to use `subscriptions`
```jsx
const Component2 = () => {
  const { subscriptions } = useIAP();

  useEffect(() => {
    // do something with subscriptions
  }, [subscriptions])
}
```

As you can see, in the above example, if using Context, we do not need to call `getSubscriptions` in `Component2`. The `subscriptions` are shared across the app. Same for other variables.

Besides, with the old implementation, `useIAP` will trigger `initConnection` and register listeners. So if I use `useIAP` in multiple places, it will call `initConnection` multiple times that will against the recommendation  in the documentation
> In order to initialize the native modules, call initConnection() early in the lifecycle of your application. This should be done at a top-level component as the library caches the native connection. Initializing just before you needed is discouraged as it incurrs on a performance hit

## To be concluded
if you are using `useIAP` hook only 1 time in entire your application, you are good. But if you'd like to use `useIAP` multiple times, the `Context` is a better choice.

Please let me know if you have any questions. Thanks

package.json
r0b0t3d: Hi @hyochan , not sure but when adding this lib from my fork. I can not find `index.js` from `node_modules/react-native-iap`. So I remove `.js` so that it can pick `index.ts` instead

hyochan (commented):
Thanks for your explaination. Agreed with your context and this can avoid the global variables either.

package.json
hyochan: @r0b0t3d Could you kindly test out after reverting this? In every `npm` tutorial, they recommend using `.js` extension. This update should be merged after we are sure with the actual consequences.

hyochan: Actually, I think you are using your own `ts` file instead of built one. you may try the `built` package using `yarn build` or `npm pack` and rename that to `react-native-iap` and move into `Example/node_modules`.

I think this would help!

r0b0t3d: I'll try then let you know. Thanks

r0b0t3d:
@hyochan I reverted the `package.json` file. Please help take a look. Thanks

hyochan (approved):
Thanks for your contribution!

PR #1498, opened 2021-09-16
Release 8.0.0

andresesfm:
- Migration to Kotlin from Java
- Migration to swift from ObjectiveC
- Fixed Amazon getProducts being called twice
- Implemented fix for Promoted product on iOS
- Removed Java style plugin
- Updated example app to RN 0.64.


hyochan (commented):
Great that we've migrate our plugin to `swift` and `kotlin`. 

I'll try to review next week. Thanks for the awesome work!

android/build.gradle
```diff
@@ -1,19 +1,19 @@
...
-  repositories {
+    ext.kotlin_version = '1.5.30'
+    repositories {
     google()
     mavenCentral()
   }
 
   dependencies {
     classpath 'com.android.tools.build:gradle:4.2.2'
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
zxcpoiu, line 11: Just a heads up.
Should we use the camel case style like: `kotlinVersion`?

Not sure if `kotlin_version` is whether sort of the android studio default or not, but I see lots of example use this one.
But in react native realm, I saw some popular repos use camel case style.

ex:

https://github.com/software-mansion/react-native-screens/blob/master/android/build.gradle#L12 
https://github.com/react-native-webview/react-native-webview/blob/master/android/build.gradle#L17
https://github.com/Nozbe/WatermelonDB/blob/master/native/android/build.gradle#L38

I'm worried I have to use both `kotlin_version` and `kotlinVersion` in my gradle ext eventually :fearful: 
Not a big deal though, thanks for the great job. :+1:

hyochan: Thanks for issuing! Looks like we need to fix to `$kotlinVersion`.

hyochan:
Closes via #1567

PR #1503, opened 2021-09-17
Improve gradle versioning for androidX to fix build error

ifsnow:
I faced the below error

```
> Task :app:checkStagingReleaseAarMetadata FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:checkStagingReleaseAarMetadata'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckAarMetadataWorkAction
   > The minCompileSdk (31) specified in a
     dependency's AAR metadata (META-INF/com/android/build/gradle/aar-metadata.properties)
     is greater than this module's compileSdkVersion (android-29).
     Dependency: androidx.browser:browser:1.4.0-alpha01.
```

I think it's good to provide the version we want to be able to set. In this modified version, a specific version can be specified.

```
buildscript {
    ext {
        ...
        androidXAnnotationVersion = "1.2.0"
        androidXBrowserVersion = "1.3.0"
    }
    ....
}
```

If `minCompileSdk` of the IAP library is not 31 or higher, it seems necessary to set the default value as follows.

```
def androidXAnnotationVersion = safeExtGet('androidXAnnotationVersion', '1.2.0')
def androidXBrowserVersion = safeExtGet('androidXBrowserVersion', '1.3.0')
```

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -75,8 +75,12 @@ dependencies {
     if (androidXVersion == null) {
       androidXVersion = defaultAndroidXVersion
     }
-    def androidXAnnotation = safeExtGet('androidXAnnotation', androidXVersion)
-    def androidXBrowser = safeExtGet('androidXBrowser', androidXVersion)
+
+    def androidXAnnotationVersion = safeExtGet('androidXAnnotationVersion', androidXVersion)
+    def androidXBrowserVersion = safeExtGet('androidXBrowserVersion', androidXVersion)
+
+    def androidXAnnotation = safeExtGet('androidXAnnotation', androidXAnnotationVersion)
+    def androidXBrowser = safeExtGet('androidXBrowser', androidXBrowserVersion)
     implementation "androidx.annotation:annotation:$androidXAnnotation"
     implementation "androidx.browser:browser:$androidXBrowser"
   }


hyochan:
@ifsnow The example `CI` is failing. Could you kindly see what is happening inside?

ifsnow:
@hyochan This is a patch to fix this build error. If `minCompileSdk` is not 31 or higher, all build will fail after clean.

hyochan:
@ifsnow Well if this is a fix, `CI` should pass in the `PR` either. @andresesfm How do you think?

andresesfm:
Hello @ifsnow can you please explain why you couldn't simply have the following in your gradle files:
```
buildscript {
    ext {
        ...
        androidXAnnotation = "1.2.0" <-- you had androidXAnnotationVersion
        androidXBrowser = "1.3.0" <-- you had androidXBrowserVersion
    }
    ....
}
```
Forgive me if I don't see your point but It looks like you want the library to adapt to your naming convention vs the other way around.

As a side note, I think the failure in CI is caused by `defaultAndroidXVersion` since it's specified with a wildcard. Looks like the latest `1.+` version requires minsdk to be 31

andresesfm:
This is the fix: https://github.com/dooboolab/react-native-iap/pull/1505

ifsnow:
@andresesfm @hyochan This is because each library has a different stable version. (androidx.annotation 1.2.0 / androidx.browser 1.3.0) Because users' environments are so diverse, I think it would be good if a way to set each version was provided.

andresesfm:
That's just the default. You can already set the value to whatever you need

Issue #1479, opened 2021-09-02
requestPurchase promise never settles when user cancels 


henninghall:
### Version of react-native-iap
7.3.0

### Version of react-native
0.64.2

### Platforms you faced the error (IOS or Android or both?)
iOS, not tested on android

### Expected behavior
`requestPurchase()` returns a promise. When the user cancels the payment I expect the promise to reject. 

### Actual behavior
When a user cancels the payment the promise from `requestPurchase` is not settled, neither resolved nor rejected.

I believe the problem lays in this conditional, preventing the rejection from happening:
https://github.com/dooboolab/react-native-iap/blob/b527d2bcd0678f42ffe80fd3e2846b58622ef226/ios/RNIapIos.m#L465

### Steps to reproduce the behavior
1. Use this code

```javascript
try {
    await requestPurchase(productId)
    // never gets here
} catch(e) {
    // never gets here
}
```
2. Cancel payment when the payment sheet is shown. Either by pressing the cancel button or touching outside the bottom sheet.

andresesfm:
I agree that this is a problem

andresesfm:
I have created a release with this fix on my fork: https://www.npmjs.com/package/@iap.dev/react-native-iap/v/7.3.1

andresesfm:
https://github.com/andresesfm/react-native-iap/commit/f0496266791a3c3118268c8a441f152aa0d3e8d2

jaaywags:
+1 Seeing this issue too

FMiscia:
Same

mvrty:
Same problem on Android (with React Native 0.65.1)

textual:
feels like this is happening to me with ios subscriptions

PR #1504, opened 2021-09-18
fix: Cancellled purchase kept promise dangling

pennersr:
If the user does not proceed with the payment, the `RNIap.requestSubscription()` promise never resolves and remains dangling. It's also an issue if you want to show a "payment in progress" spinner while the user is interacting with the system level payment flows, as you don't get informed about the user cancelling, meaning, the spinner spins forever.  

Fixes: https://github.com/dooboolab/react-native-iap/issues/1479

### DIFF
--- a/android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
+++ b/android/src/play/java/com/dooboolab/RNIap/RNIapModule.java
@@ -583,9 +583,7 @@ public class RNIapModule extends ReactContextBaseJavaModule implements Purchases
       error.putString("message", errorData[1]);
       sendEvent(reactContext, "purchase-error", error);
 
-      if (responseCode != BillingClient.BillingResponseCode.USER_CANCELED) {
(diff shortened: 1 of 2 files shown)


pennersr:
Note that the build failure is not due to my changes but due to upstream dependencies (minCompileSdk 31).

andresesfm:
Thank you for your contribution @pennersr. I have implemented this same fix here: https://github.com/dooboolab/react-native-iap/pull/1498 this change will be overridden by the conversion to Kotiln.
Having said that, I agree that it is the right approach

andresesfm:
On second thought, the other PR might take a while to review and this can be an interim fix. While kotlin lands. It LGTM and the CI fail is fixed here: https://github.com/dooboolab/react-native-iap/pull/1505

andresesfm (approved):

hyochan (approved):
LGTM!

hyochan (approved):
Agree with @andresesfm. Let's ship this `PR` first since we need time to review.
If possible, @pennersr, you may also try out #1498 and help us reviewing :pray:

PR #1505, opened 2021-09-19
Setting default AndroidX version

andresesfm:
It is now set to the last known working version. Note this will still fail CI until there's a new release and IapExample's package version is pointed to that new version

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -71,7 +71,7 @@ dependencies {
     implementation "com.android.support:support-annotations:$supportLibVersion"
     implementation "com.android.support:customtabs:$supportLibVersion"
   } else {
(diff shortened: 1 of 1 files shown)


hyochan (approved):
Thanks for the fix!

I'll add @ifsnow as a co-author for current `PR` as a first issuer :pray:

angelos3lex:
@hyochan We could also specify the version as "1.3+" and it would still build.  Any reason for going back to 1.2 instead?

andresesfm:
@angelos3lex According to this: https://developer.android.com/jetpack/androidx/versions for annotations 1.3.0 is still in alpha. 1.2.0 is safe for the local build. You can change in your project by setting the value in `ext`

PR #1521, opened 2021-09-28
fix countryCode typing in ProductCommon interface

yagovv:
This fixes a type in the ProductCommon interface. 
As long as I have tested in IOS, the ProductCommon does not receive `countryCodeIOS` but `countryCode` and none in Android.


andresesfm:
LGTM

andresesfm (approved):

PR #1524, opened 2021-10-04
fix NativeEventEmitter warnings since ReactNative 0.65

sbin1211:


radko93:
@andresesfm can we edit the PR? It's not great that we cannot get it in just because of the spaces.

PR #1525, opened 2021-10-04
Using TCK Tested JDK builds of OpenJDK

carldea:
The AdoptOpenJDK has been discontinued since July 2021 (https://adoptopenjdk.net). Switching the distribution to Azul Zulu. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. 

Also added are fixed (major) release version(s) such as `JDK 11.0.3`. This is often a good practice whenever a build/test triggers (push/pull events) to help determine if the latest (`JDK 11`) had failed the from the **latest build** vs something in **your code** caused (or introduced). 

**For example**, when building with `JDK 11.0.3` (fixed version) the build/tests passes (Green) and JDK 11 fails (Red) will mean that the latest `JDK 11` was the cause and not your code. 

**Note:** Other distributions such as Temurin do not support archived fixed releases prior to Sept. 2021 and many non LTS (long term support) releases if you plan to try out newer features in the language.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,14 +25,16 @@ jobs:
   build_android:
 
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        java-version: [ 11.0.3, 11 ]
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 11
+    - name: Set up JDK ${{ matrix.java-version }} 
       uses: actions/setup-java@v2
       with:
-        distribution: 'adopt'
-        java-version: '11'
+        distribution: 'zulu'
+        java-version: ${{ matrix.java-version }}
     - name: Grant execute permission for gradlew
       run: chmod +x android/gradlew
     - name: Build with Library with Gradle and check lints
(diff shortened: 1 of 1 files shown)


carldea:
Full disclosure, I work for Azul and here are my opinions:

**Paid support**
I believe both are fine (both distros are derived from OpenJDK and TCK tested), the only real difference is AWS customers using Corretto may later want Amazon paid support  for apps running in production environments from scaling, tuning, monitoring services, etc. Zulu is also free and if a customer or ISV(independent software vendor) later want paid support they’d choose Azul. 

**As mentioned earlier about build distro differences:**

Corretto doesn’t support every OpenJDK version (non-LTS or fixed versions). Azul Zulu does🙂.
Zulu offers Java 7-18ea (latest) and archived versions (fixed major releases). 

**Repo owner’s perspective:**
As a project/repo that is a library or framework it is a good idea to stick with a JDK distribution such as Zulu because it’s free, up to date, TCK tested (unlike AdoptOpenJDK), consistent track record & has every version of OpenJDK. 
Originally GitHub action `setup-java@v1` only used `zulu` as a distro. With v2 it’s now a requirement to choose a distro. That’s when lots of builds broke or weren’t getting the latest (patched) builds. The GH Action template continues to have `adopt`.

I hope that makes sense.
-Carl

carldea:
> 👍

Thank you. Glad to help contribute!

PR #1542, opened 2021-10-17
Fix Typescript: No index signature with a parameter of type 'string' was found on type

albertovelazmoliner:
Running yarn build
`$ yarn tsc`

I'm getting this error:

 `node_modules/react-native-iap/src/iap.ts:118:22 - error TS7053: Element implicitly has an 'any' type because expression of type 'any' can't be used to index type '{ CA: string; ES: string; AU: string; DE: string; IN: string; US: string; JP: string; GB: string; IT: string; BR: string; FR: string; }'`

`118     const currency = currencies[user.userMarketplaceAmazon];`

This PR fix that issue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,14 +25,16 @@ jobs:
   build_android:
 
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        java-version: [ 11.0.3, 11 ]
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


andresesfm:
Hello @albertovelazmoliner I'm curious to know how you are getting that error. CI runs `yarn build` that in turn runs `tsc` and I don't see it failing with that error. So I wanted to ask you about the version of the library you are running this on and what's different from what I just mentioned?

PR #1544, opened 2021-10-19
fix: NativeEventEmitter warnings since ReactNative 0.65

radko93:
closes #1524. Superseeds other PR as it also includes the change for `RNIapModule.java`


hyochan (approved):
LGTM!

Issue #1562, opened 2021-11-09
Long lived connections stale/become unusable


ospfranco:
### Version of react-native-iap
7.5.0

### Version of react-native
0.65.1

### Platforms you faced the error (IOS or Android or both?)
Android

Hi, we migrated from 6.2.4 to 7.5.0 a month ago, from that moment our error rates started sky-rocketing, we have been trying to debug the issues for some days now, but we couldn't find any issue with our code but there is nothing wrong, however, we did notice somethings:

- We initialize the connection once on app start, and we let the object alive through the app's lifecycle
- On the 6.3.0 release, the logic changed to keep only one client alive (https://github.com/dooboolab/react-native-iap/commit/1409b8f13cb4d3eb60b4c5eac17bab8edaf3cd7d)
- We think a combination of us keeping the billingClient alive for a long period of time + now the handling of a single client instance, causes the client at some point to disconnect and never to reconnect again, causing elevated error rates

Any ideas if this is correct or if something else could be going on?

jeremyhalin:
Hi @ospfranco 

I face the same problem as you (I mentioned my issue that is, I think, related).
What did you do to solve this? Did you downgrde to `^6.0`?

I'm using hooks, do you?

ospfranco:
No, I'm using redux sagas, I added a new method (android only) isReady, if that returns false, you can kill and re-initialize the connection

jeremyhalin:
>No, I'm using redux sagas, I added a new method (android only) isReady, if that returns false, you can kill and re-initialize the connection

Ok thanks, I don't quite know where to add the call of `isReady` in my situation since I'm using hooks.

>you can kill and re-initialize the connection

How do I do that?

In my root container
```javascript
const {
  currentPurchase,
  finishTransaction,
} = useIAP();

useEffect(() => {
  const checkCurrentPurchase = async (purchase): Promise<void> => {
    if (purchase) {
      const receipt = purchase.transactionReceipt;
      if (receipt)
        try {
          await finishTransaction(purchase);
          await dispatch(Purchase.action(purchase))
          // display something
          showMessage({
            message: t('shop.successMessage.message'),
            description: t('shop.successMessage.description'),
            type: 'success',
          })
        } catch (ackErr) {
          console.warn('ackErr', ackErr);
          showMessage({
            message: ackErr.message,
            type: 'danger'
          })
        }
      }
  };

  checkCurrentPurchase(currentPurchase);
[...]

Issue #1559, opened 2021-11-08
Question/Suggestion: Check connection state


ospfranco:
Hi, thanks a lot for the package!

We are facing some issues, we initialize the connection on app start because we have a lot of upsell dialogs and other user prompts to trigger payment/subscription. The problem is, however, that since we open the connection on app start and we need to keep it open if the app goes into the background, it appears that we lose the connection at different points and payments start behaving erratically.

Is there any way to query for the connection state and/or close it programmatically? is this a valid use case, of the connection should always remain open once the initConnection method has been called?

mifi:
Maybe this can be closed because the functionality was added?

PR #1563, opened 2021-11-09
Expose new isReady method

ospfranco:
Related to #1562 and #1559


enricsa (approved):

hyochan (changes requested):
Could you kindly rename `isReady` to `isReadyAndroid` if this is only `android` dependant?

ospfranco:
There seems to be a linting issue, but I cannot find anything wrong, the actions output is also not clear... any pointers?

ospfranco:
Ah, nvm, found the linting issue, I need to run `./gradlew googleJavaFormat`

hyochan (approved):
LGTM!

Dallas62:
`undefined is not a function` Not sure the method is correctly exposed...

Dallas62:
I guess something like:
```js
export const isReadyAndroid = (): Promise<boolean> =>
  getNativeModule(). isReadyAndroid();
```
is missing in the `src/iap.ts`

ospfranco:
You are right, here is a new PR, please do check it carefully since I don't understand all the internals of the package

https://github.com/dooboolab/react-native-iap/pull/1604/files

PR #1566, opened 2021-11-12
Refresh SkuDetails

rrrasti:
In the current Android implementation is it possible to get the following error during the invocation of `RNIap.requestPurchase(...)`:
```
{
	message: 'Google is indicating that we have some issue connecting to payment.',
	debugMessage: 'Invalid SKU details.',
	code: 'E_DEVELOPER_ERROR',
	responseCode: 5
}
```
It appears that the problem is caused by the changing content of `SkuDetails` (e. g. `skuDetailsToken` in `originalJson`) over time even for the same product. It can happen that after several invocations of `RNIap.getProducts(...)` there will be multiple instances of `SkuDetails` in the `skus` list with the same `sku/productId` and later a wrong instance is chosen to purchase a product during the invocation of `RNIap.requestPurchase(...)`. The proposed implementation should minimize aforementioned error.


hyochan (approved):
Very strange stuff. However, the code looks better.

PR #1568, opened 2021-11-15
[8.0] android: remove deprecated method createJSModules in module package

zxcpoiu:
RN 0.47 deprecated the method. Remove it entirely.

Some may just remove the `override` to pass the build (`createJSModules overrides nothing` error), but this function is actually can be removed entirely.

see: https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8


hyochan (commented):
Thanks for issuing this!

Do you think we should change the `peerDependencies` to `>=0.60` when merging this `PR`?

hyochan (approved):

PR #1578, opened 2021-11-26
fix: android querySkuDetailsAsync ui thread

FallenDestiny:
In Billing 4 response from `querySkuDetailsAsync` is returned in background, it causes android app in production to get stuck at this point and it makes impossible further purchases or any other interactions with SKUs. Wrapping `querySkuDetailsAsync` handler in `runOnUiThread` fixes the issue


FallenDestiny:
@hyochan Can we please make sure this also gets fixed in v8 kotlin code

hyochan:
@FallenDestiny Hello! Thanks for your contribution. I currently don't have a good test env on my machine but I'll give it a try.

Also Please check the CI failing :pray:

hyochan (approved):
Thanks for your contribution!

I will try to apply code on `kotlin`. Maybe @andresesfm  could help either.

Issue #1571, opened 2021-11-17
Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported`


tonsnoei:
> Please use [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want get some help out of it. Please report issue if there is a bug.

### Version of react-native-iap
7.5.1

### Version of react-native
0.66.2

### Platforms you faced the error (IOS or Android or both?)
Android

### Expected behavior
No error

### Actual behavior
Following error when running ./gradlew assembleRelease

```
> Task :react-native-iap:processAmazonReleaseManifest FAILED
/Users/ton.snoei/Documents/dev.nosync/trilab/IEXv7/src/reactnative/node_modules/react-native-iap/android/src/main/AndroidManifest.xml Error:
        Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.

See http://g.co/androidstudio/manifest-merger for more information about the manifest merger.
```

### Tested environment (Emulator? Real Device?)
n/a

### Steps to reproduce the behavior
Use the following app/build.gradle settings
```
        buildToolsVersion = "30.0.2"
        minSdkVersion = 21
        compileSdkVersion = 31
        targetSdkVersion = 31
```

gollobo:
I have the same problem, apparently the AndroidManifest.xml inside `/react-native-iap/android/src/amazon/` doesn’t have the `android:exported` value in the `<receiver/>` but I don’t know the correct value

tonsnoei:
You can resolve it yourself by replace the contents of the file `/node_modules/react-native-iap/android/src/amazon/AndroidManifest.xml` with

```xml
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
          package="com.dooboolab.RNIap">
<application android:exported="true">
    <receiver android:name = "com.amazon.device.iap.ResponseReceiver"
        android:permission = "com.amazon.inapp.purchasing.Permission.NOTIFY" 
        android:exported="true">
        <intent-filter>
            <action android:name = "com.amazon.inapp.purchasing.NOTIFY" />
        </intent-filter>
    </receiver>
</application>
</manifest>
```

And the file `/node_modules/react-native-iap/android/src/main/AndroidManifest.xml`
```xml
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
          android:exported="true"
[...]

PR #1586, opened 2021-12-05
Added exported variable in main and amazon manifest files

SergiOnGit:


android/src/amazon/AndroidManifest.xml
```diff
@@ -1,13 +1,12 @@
-
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="com.dooboolab.RNIap">
-<application>
+<application android:exported="true">
```
hyochan, line 3: @SergiOnGit Could you kindly explain how this fixes the compatibility issue? As far as I know this just allows another app to start the activity.

SergiOnGit: @hyochan Sorry, looks like I wrote another library's PR title here.

SergiOnGit: @hyochan Can you accept this PR? This is really important.

hyochan: I am sorry for the late review. @SergiOnGit Could you kindly explain how this is important in `amazon` IAP?

Dallas62: @hyochan This is blocking build on Android, don't if should be exported true or false, but it's blocking. #1571 

```bash
> Task :react-native-iap:processAmazonReleaseManifest FAILED
/Users/boristacyniak/Projects/Mobile/Monstera/node_modules/react-native-iap/android/src/main/AndroidManifest.xml Error:
        android:exported needs to be explicitly specified for <receiver>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.

```

SergiOnGit: > @SergiOnGit Could you kindly explain how this fixes the compatibility issue? As far as I know this just allows another app to start the activity.

I don't know native code, I just know that without this app doesn't build. Tell you the truth it is very bad that this kind of good module have this bug for this long. This doesn't happen on windows only, I have tested on mac too and it's the same.

wandersonalwes:
Please approve.

wandersonalwes:
Without that, I'm having conflict to compile an app after updating the `buildscript`.

```
buildscript {
    ext {
        buildToolsVersion = "30.0.2"
        minSdkVersion = 23
        compileSdkVersion = 31
        targetSdkVersion = 31
        ndkVersion = "21.4.7075529"
        androidXAnnotation = "1.1.0"
        androidXBrowser = "1.0.0"
    }
}
```

hyochan (approved):
Sorry for the late review~!

I think there is enough detail to be merged.

Thanks for the fixes 🙏

PR #1604, opened 2022-01-03
Fix missing `isReadyAndroid` method in iap.ts

ospfranco:
Forgot to expose the method on the main ts file, please re-check and let me know if anything else is missing (since I don't know the internal workings of the package)


hyochan (approved):
Thanks for the fixes!

PR #1607, opened 2022-01-06
ignore yarn cache in npmignore

effektsvk:
Hello,

when installing your package I noticed that its size was over 200 MB. When checking package files, I found out that `.yarn/cache` folder takes those 200 MB. I tried to pack with the updated `.npmignore` and the package size is just 1.0 MB.

You probably accidentally included it in published package, it is ignored in `.gitignore` but not in `.npmignore`.

image

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -21,3 +21,4 @@ docs/
 .eslintignore
 .prettierrc.js
 .gitattributes
+.yarn/cache


hyochan (approved):
OMG! Thanks for the catch 💯 

I forgot to ignore this after upgrading pkg manager to yarn berry 😞

Issue #1555, opened 2021-11-05
8.0.0 RC3 : buyProduct:andDangerouslyFinishTransactionAutomatically:resolve:reject: is not a recognised Objective-C method


crohr:
### Version of react-native-iap

8.0.0-rc.3

### Version of react-native

0.64.0

### Platforms you faced the error (IOS or Android or both?)

iOS

### Expected behavior

Can actually request a subscription.

### Actual behavior

Crash with error (see screenshot for full trace):

```
Exception 'buyProduct:andDangerouslyFinishTransactionAutomatically:resolve:reject: is not a recognised Objective-C method.' was thrown while invoking `buyProduct on target RNIapIos with params ...
```

![IMG_B23BC9878BFE-1 3](https://user-images.githubusercontent.com/6114/140520609-bb633054-5015-4505-a5af-93b466f30820.jpeg)

### Tested environment (Emulator? Real Device?)

Emulator and real device

### Steps to reproduce the behavior

Setup IAP according to doc for 8.0.0, then:

```
requestSubscription('product_id', false) // same error when giving true for andDangerouslyFinishTransactionAutomaticallyIOS
```

Might be interesting for @andresesfm, which, as far as I understand, is the one developing 8.0.0?

andresesfm:
Thank you @crohr I'll take a look

lunarraid:
I've been running into this as well. Would downgrading be too much of a challenge at this point? I didn't realize at the time of install this was still only a preview.

jspark-qs:
I had same issue
I tried to fix

andresesfm:
Implemented suggestion by @jspark-qs here: https://github.com/dooboolab/react-native-iap/pull/1627

PR #1627, opened 2022-01-24
fix paramenter name

andresesfm:
Implements fix proposed here: https://github.com/dooboolab/react-native-iap/issues/1555


hyochan (approved):
LGTM!

PR #1652, opened 2022-02-25
add kotlin linter

andresesfm:
Adds Ktlint (https://github.com/pinterest/ktlint) as a step to CI and committing the formatted Kotlin files


hyochan (approved):
Awesome thing again 💯

PR #1678, opened 2022-04-11
fix: minimum version for tvOS

### LINKED ISSUE #1677: Compilation error for tvOS: 'init(identifier:keyIdentifier:nonce:signature:timestamp:)' is only available in tvOS 12.2 or newer
> Please use [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want get some help out of it. Please report issue if there is a bug.

8.0.7

### Version of react-native

npm:react-native-tvos@0.66.3-1

tvOS

### Expected behavior

Compile successfully for tvOS

### Actual behavior

Compilation error in 54 different places where the `#available` attribute is used to check the platform version. All errors follow this format:

```
'init(identifier:keyIdentifier:nonce:signature:timestamp:)' is only available in tvOS 12.2 or newer
```

Device

1. Create react-native project
2. Add `react-native-iap` as a dependency
3. Try to run the project in an Apple TV

fontesrp:
Fixes #1677

Add check for tvOS version where the `#available` attribute is used

### DIFF
--- a/ios/RNIapIos.swift
+++ b/ios/RNIapIos.swift
@@ -236,8 +236,8 @@ class RNIapIos: RCTEventEmitter, SKRequestDelegate, SKPaymentTransactionObserver
             addPromise(forKey: prod.productIdentifier, resolve: resolve, reject: reject)
             
             let payment: SKMutablePayment = SKMutablePayment(product: prod)
-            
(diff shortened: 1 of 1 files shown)


hyochan (approved):
LGTM!

PR #1704, opened 2022-05-27
Avoid finishing transactions that can still succeed and have not been consumed

posuna-cameo:
Recoverable errors like interrupted purchases, network failures, failures while processing receipts, etc, are being consumed even though they were recoverable.
Clear transactions calls finishTransaction, which ends updates, and therefore shouldn't be called while there are recoverable transactions awaiting updates.

### DIFF
--- a/IapExample/App.js
+++ b/IapExample/App.js
@@ -118,6 +118,12 @@ class Page extends Component {
       await RNIap.initConnection();
       if (Platform.OS === 'android') {
         await RNIap.flushFailedPurchasesCachedAsPendingAndroid();
+      } else {
+        // WARNING This line should not be included in production code
+        // This call will call finishTransaction in all pending purchases on every launch,
+        // effectively consuming purchases that you might not have verified the receipt or given the consumer their product
+        // TD:DR you will no longer receive any updates from Apple on every launch for pending purchases
+        await RNIap.clearTransactionIOS();
       }
     } catch (err) {
       console.warn(err.code, err.message);
--- a/IapExample/App.js
+++ b/IapExample/App.js
@@ -118,8 +118,6 @@ class Page extends Component {
       await RNIap.initConnection();
       if (Platform.OS === 'android') {
         await RNIap.flushFailedPurchasesCachedAsPendingAndroid();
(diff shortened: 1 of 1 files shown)


IapExample/App.js
```diff
@@ -118,8 +118,6 @@ class Page extends Component {
       await RNIap.initConnection();
       if (Platform.OS === 'android') {
         await RNIap.flushFailedPurchasesCachedAsPendingAndroid();
-      } else {
-        await RNIap.clearTransactionIOS();
```
hyochan, line 122: If I remember correctly, this was to help people test consumable purchases. Without this, they should implement checking receipts and then clear transactions. 

Do you still think this should be removed in the example project?

posuna-cameo: I do think it should be removed. People who use this code as base to implement consumable purchases, will find their purchases being cleared for them on relaunch without their knowledge, and won't understand why they don't receive updates for their purchases.
At the very least, a comment should be added that this call will consume all pending transactions, and that they won't get updates on consecutive launches. But I still think removing it is the better option.

hyochan: Thanks for the reply! Could you kindly make some comments without just removing all?
I think people might have questions about how to repurchase consumable products. We used to have many questions on that made us have this code in the example.

IapExample/App.js
```diff
@@ -119,6 +119,10 @@ class Page extends Component {
       if (Platform.OS === 'android') {
         await RNIap.flushFailedPurchasesCachedAsPendingAndroid();
       } else {
+        // WARNING This line should not be included in production code
+        // This call will call finishTransaction in all pending purchases on every launch,
+        // effectively consuming purchases that you might not have verified the receipt or given the consumer their product
+        // TD:DR you will no longer receive any updates from Apple on every launch for pending purchases
```
hyochan, line 125: What is TD:DR? I've only heard about TL;DR so sorry if I don't understand this 🙏

posuna-cameo: It's a typo :)

hyochan (approved):
LGTM!

Issue #1687, opened 2022-05-03
Android compile error on RN 0.68: Could not find method compile() for arguments [project ':react-native-iap'] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.


mcarlson:
> Please use [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want get some help out of it. Please report issue if there is a bug.

### Version of react-native-iap
I tried 6.2.4 (what we're on) all the way up to the latest: 8.0.8

### Version of react-native
0.68.1

### Platforms you faced the error (IOS or Android or both?)
Android

### Expected behavior
Successful compile

### Actual behavior
Gradle 7 fails with:
`Could not find method compile() for arguments [project ':react-native-iap'] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.`

### Tested environment (Emulator? Real Device?)
Emulator

### Steps to reproduce the behavior
Update to RN 0.68 and attempt to compile on Android. 

Apparently, compile() was deprecated in Gradle since version 4 and has now been removed in Gradle 7. You're meant to use implementation(), see https://stackoverflow.com/questions/58594801/could-not-find-method-compile-for-arguments-com-android-supportappcompat-v7 and https://stackoverflow.com/questions/70257358/could-not-find-method-compile-for-arguments-project-react-native-google-places for more details.

PR #1714, opened 2022-06-06
Fix documentation to use implementation instead of compile (android)

andresesfm:
Fixes documentation as mentioned here https://github.com/dooboolab/react-native-iap/issues/1687


hyochan (approved):
Thanks :)

PR #1718, opened 2022-06-07
Minor typo/grammar change.

RussVanBert:

PR #1719, opened 2022-06-08
Remove legacy method isReadyAndroid()

andresesfm:
The `isReadyAndroid` is a Legacy method that comes from the 7.x times. It is both broken (removed from native code before 8.0.0) and redundant since `initConnection` handles duplicate attempts

- Closes #1718

Co-authored-by:@RussVanBert <russ_vb@yahoo.com.au>


andresesfm:
The JS method was introduced here: https://github.com/dooboolab/react-native-iap/commit/10b70827aee78f98ec32846f53c4da1d3ea1b7fd by @ospfranco But the native method was removed. As I mentioned in the description this is no longer needed.
This: https://github.com/dooboolab/react-native-iap/blob/main/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt#L61 handles the check with no performance penalty.

Please let me know if any concerns

ospfranco:
I think I already went over this in one of the issues, I have no idea why the native method was removed. The problem is not duplicated initialization attempts, but rather that the connection can silently die while the app is backgrounded and there is no way to check if re-initialization is needed. At least that was the case, when I created the PR, if that is no longer the case, then sure you can remove it. Otherwise, you still need some way to query the state of the client for long-lived instances (not in a component, but in a redux store for example, which was our case)

hyochan:
> ospfranco

`isReadyAndroid` can be achived with `initConnection` in `idle` case, https://github.com/dooboolab/react-native-iap/blob/60a2b23794224ba06079119f1a1db8023f5b9e86/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt#L60-L69.

Is there any case where it'd be not enough?

hyochan (approved):
Let's merge this for now since the method itself malfunctioned.

Future discussion is always welcomed!

Issue #1717, opened 2022-06-07
Hard crash on iOS calling getSubscriptions() for subscription with trial / promotional offer


bitfabrikken:
> Please use [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want get some help out of it. Please report issue if there is a bug.

### Version of react-native-iap
8.0.10

### Version of react-native
0.67.2

### Platforms you faced the error (IOS or Android or both?)
IOS

### Expected behavior
Call getSubscriptions() without crashing

### Actual behavior
Calling getSubscriptions() crashes the app.

### Tested environment (Emulator? Real Device?)
Both

### Steps to reproduce the behavior
Call getSubscriptions() within an app that's live on the App Store, which has a subscription with a promotional offer for a trial of 1 week.

I call this:
```
await RNIap.getSubscriptions(["supporter"]);
```
Then the console in Xcode says "Add new object : supporter", and crash occurs.
Here's a crash log from device: https://pastebin.com/JvXBmzqz
Let me know if anything else can help.

andresesfm:
Hello @bitfabrikken Looking at the logs. Seems like this is the line where it blows up: https://github.com/dooboolab/react-native-iap/blob/main/ios/RNIapIos.swift#L776 Could it be that the discount has a strange price Locale?

andresesfm:
Doesn't look like it can be null or have a strange value: https://developer.apple.com/documentation/storekit/skproduct/3043534-discounts

andresesfm:
This is my proposed solution: https://github.com/dooboolab/react-native-iap/pull/1720

hyochan:
Please come back if this happens in `8.1.0`.

hyochan:
Reopened because this is reverted in `8.1.1`.

sampsonjoliver:
We are also seeing this crash in `v8.6.5`

```
EXC_BREAKPOINT: 

0  libswiftFoundation.dylib static Locale._unconditionallyBridgeFromObjectiveC(_:)
1  App          specialized RNIapIos.getDiscountData(_:) (RNIapIos.swift:795:37)
2  App          RNIapIos.getDiscountData(_:) (<compiler-generated>)
3  App          RNIapIos.getProductObject(_:) (<compiler-generated>)
4  App          thunk for @callee_guaranteed () -> () (<compiler-generated>)
5  App          thunk for @escaping @callee_guaranteed () -> () (<compiler-generated>)
[...]

PR #1720, opened 2022-06-08
swift guard against possible nil value

andresesfm:
Attempt to fix: https://github.com/dooboolab/react-native-iap/issues/1717

Not a swift expert but it seems like this would work. I found that we have to check for possibly nil even if the type is not optional as per https://stackoverflow.com/a/44060466/570612


hyochan (approved):
Looks like this is a good attempt.
This is strange though since the [document](https://developer.apple.com/documentation/storekit/skproduct/3043534-discounts) doesn't tell that it can be optional as you've pointed out.

Thank you for finding out about all this.

PR #1727, opened 2022-06-17
Refactored ensure connection to reconnect

andresesfm:
This is my proposed change to attempt a reconnection if the billing client is not initialized. It looks more dramatic than it actually is. The only real change was on `ensureConnection`. All the other changes are formating caused by removing the custom callback `EnsureConnectionCallback` and using lambdas instead since we no longer need to pass the billing Client around as a parameter


hyochan (commented):
Looks like I need to spend some time reviewing this.
If I remember correctly, I think @LinusU can help review this since `ensureConnection` was his first proposal.

@LinusU If you are available I hope you can help us with this `PR` 🙏

hyochan (approved):
Looks good to me so far.

Tested on my [environment](https://github.com/dooboolab/dooboolab.com)

ifsnow:
@andresesfm That's a great improvement!! Thanks

PR #1730, opened 2022-06-20
android unit tests

andresesfm:
- Refactored ensure connection to reconnect
- ktlint 4.10.0
- added import
- Unit tests for initConnection
- Reverts local change

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -50,6 +50,9 @@ android {
       dimension "store"
     }
   }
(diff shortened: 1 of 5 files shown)


hyochan (commented):
@andresesfm This is really great!
Could you also add this to `CI` step?

Looks like android unit test is not covered in CI.

hyochan (approved):
Oh yes! You are right 👍

PR #1731, opened 2022-06-22
test flush failed purchases cache

andresesfm:


android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -142,11 +138,8 @@ class RNIapModule(
                                 )
                             return@ConsumeResponseListener
                         }
-                        try {
-                            promise.resolve(true)
-                        } catch (oce: ObjectAlreadyConsumedException) {
-                            promise.reject(oce.message)
```
andresesfm, line 148: This would be a crash

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -81,30 +80,27 @@ class RNIapModule(
...
-            promise.resolve(true)
+            promise.safeResolve(true)
             return
         }
         if (googleApiAvailability.isGooglePlayServicesAvailable(reactContext)
             != ConnectionResult.SUCCESS
         ) {
             Log.i(TAG, "Google Play Services are not available on this device")
-            promise.reject(DoobooUtils.E_NOT_PREPARED, "Google Play Services are not available on this device")
+            promise.safeReject(DoobooUtils.E_NOT_PREPARED, "Google Play Services are not available on this device")
```
hyochan, line 90: Could you pin me on the link where I can learn more about `safeReject` and `safeResolve`?

I want to be more specific about when to use these 🤔

hyochan: @andresesfm Looks like the method does not exist anymore in [react-native repo, Promise.java](https://github.com/facebook/react-native/blob/main/ReactAndroid/src/main/java/com/facebook/react/bridge/Promise.java).

andresesfm: It's a kotlin extension method I created in PromiseUtils. It's more convenient so se don't have the try-catch block for promises everywhere

hyochan: Oh. It looks cool!

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
hyochan: The user would use the `try` and `catch` block on the JS side when it is a promise. Then it won't crash.

But I think we can accept this since it doesn't look like a big deal to me.

hyochan (approved):
Thanks the contribution again 🙏

Issue #1733, opened 2022-06-23
Reconnecting billingClient is not working


Mkolakoglu:
> Please use [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want get some help out of it. Please report issue if there is a bug.

### Version of react-native-iap
8.1.3
### Version of react-native
0.68.1
### Platforms you faced the error (IOS or Android or both?)
Android
### Expected behavior
When I call `initConnection` -> `endConnection` -> `initConnection` (again), the iap service can not be used. I am thinking that it will not reconnect when the connection is lost because of the long lived connections on Android.
### Actual behavior
The iap service should be able to be used after calling `initConnection` -> `endConnection` -> `initConnection` (again)
### Tested environment (Emulator? Real Device?)
Real Device
### Steps to reproduce the behavior
Call `initConnection` -> `endConnection` -> `initConnection` and then try to purchase subscription

andresesfm:
@Mkolakoglu can you please describe if you get an error? are you manually ending the connection every time? As you can see here: https://github.com/android/play-billing-samples/blob/master/TrivialDriveJava/app/src/main/java/com/sample/android/trivialdrivesample/billing/BillingDataSource.java#L206 They don't need to set the client to null, they simply retry to connect.  

@ifsnow you mentioned you are experiencing the same even after my changes? If so, are you saying that @Mkolakoglu 's PR fixes it for you?

Note that I have no problem with setting the billing client to null, I just think it's better to follow the official example from Google but if that doesn't work we can modify it

ifsnow:
@Mkolakoglu This issue occurs in the latest version. Rolling back to the previous version fixed it right away.

You can easily reproduce it with the steps below. It seems to be the same in the emulator.

- Start the app
- Using react-native-iap
- Close the app with the back button
- Restart the app
- Connection error when using react-native-iap

andresesfm:
@ifsnow Do you mean rolling back from `8.1.3` back to `8.1.2` fixed the issue?

andresesfm:
Can you kindly post here any error messages/logs you are seeing? It might be easier to rollback that small change than it is to convert to null objects

Mkolakoglu:
Here is the error log: 

<img width="1039" alt="Screen Shot 2022-06-24 at 10 47 20" src="https://user-images.githubusercontent.com/16003411/175499452-5256063e-526a-4c0c-b69c-79db5c1b606e.png">

[...]

PR #1734, opened 2022-06-23
fixed reconnection after losing the connection with the billingClient

Mkolakoglu:
Fixing #1733 

Please review carefully. It has been quite a while since I developed Android project 👀

### DIFF
--- a/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
+++ b/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
@@ -40,15 +40,15 @@ class RNIapModule(
     ReactContextBaseJavaModule(reactContext),
     PurchasesUpdatedListener {
 
-    private var billingClient: BillingClient? = null
-
+    private val billingClient: BillingClient = builder.setListener(this)
+        .build()
     private val skus: MutableMap<String, SkuDetails> = mutableMapOf()
     override fun getName(): String {
         return "RNIapModule"
     }
 
     private fun ensureConnection(promise: Promise, callback: () -> Unit) {
-        if (billingClient?.isReady == true) {
+        if (billingClient.isReady) {
             callback()
             return
         } else {
@@ -76,7 +76,7 @@ class RNIapModule(
 
     @ReactMethod
     fun initConnection(promise: Promise) {
-        if (billingClient?.isReady == true) {
+        if (billingClient.isReady) {
             Log.i(
                 TAG,
                 "Already initialized, you should only call initConnection() once when your app starts"
@@ -92,10 +92,7 @@ class RNIapModule(
             return
         }
 
-        billingClient = BillingClient.newBuilder(reactContext).enablePendingPurchases().setListener(this)
-            .build()
-
-        billingClient?.startConnection(
+        billingClient.startConnection(
             object : BillingClientStateListener {
                 override fun onBillingSetupFinished(billingResult: BillingResult) {
                     val responseCode = billingResult.responseCode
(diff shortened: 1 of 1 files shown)


ifsnow:
@Mkolakoglu Thanks for your work!

I also faced this problem on real devices(Android OS 8 and lower).

The reproduction scenario is as follows.
- Start the app
- Using react-native-iap
- Close the app with the back button
- Restart the app
- Connection error when using react-native-iap

I also guess that there is a problem in the part that @Mkolakoglu modified.

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -40,15 +40,15 @@ class RNIapModule(
     ReactContextBaseJavaModule(reactContext),
     PurchasesUpdatedListener {
 
-    private val billingClient: BillingClient = builder.setListener(this)
-        .build()
+    private var billingClient: BillingClient? = null
```
andresesfm, line 43: Why do we need to set this to null? that's not what the samples in the billingCllient lib suggest

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -92,7 +92,10 @@ class RNIapModule(
             return
         }
 
-        billingClient.startConnection(
+        billingClient = BillingClient.newBuilder(reactContext).enablePendingPurchases().setListener(this)
```
andresesfm, line 95: This was moved up so BillingClient can be mocked for unit tests

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -109,15 +112,32 @@ class RNIapModule(
                 }
 
                 override fun onBillingServiceDisconnected() {
-                    Log.i(TAG, "Billing service disconnected")
+                    try {
+                        billingClient = null
```
andresesfm, line 116: not needed

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -109,15 +112,32 @@ class RNIapModule(
...
             })
     }
 
     @ReactMethod
     fun endConnection(promise: Promise) {
-        billingClient.endConnection()
-        promise.resolve(true)
+        if (billingClient != null) {
+            billingClient = try {
+                billingClient!!.endConnection()
```
andresesfm, line 129: This can lead to NPE

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -453,10 +473,6 @@ class RNIapModule(
                 val subscriptionUpdateParams = subscriptionUpdateParamsBuilder.build()
                 builder.setSubscriptionUpdateParams(subscriptionUpdateParams)
             }
-            val flowParams = builder.build()
```
andresesfm, line 456: why remove this?

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
Mkolakoglu: I see, but if we don't create a new instance here, it's causing this error:
Screen Shot 2022-06-24 at 10 47 20

The rest can be reverted but I guess, we should keep this change.

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
Mkolakoglu: It seems to be not used anymore

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
Mkolakoglu: Reverted 👍

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
Mkolakoglu: Reverted 👍

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
Mkolakoglu: I reverted this anyway 👍

Mkolakoglu:
@andresesfm Thanks for your review!

I reverted the redundant changes. Please take a look again

andresesfm (approved):

andresesfm:
That works for me. Thank you @Mkolakoglu. I have approved it.

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -92,6 +91,8 @@ class RNIapModule(
             return
         }
 
+        billingClient = builder.setListener(this).build()
```
hyochan, line 94: Shouldn't we check this conditionally?
It'd be safer if we can set the listener only when the client is `closed`.

Mkolakoglu: Can we not consider as the client is closed when `billingClient.isReady == false`? We are checking this at the beginning of `initConnection`

hyochan: @Mkolakoglu Thanks for correcting the conversation. You're right 👍

hyochan (approved):
LGTM!

PR #1740, opened 2022-07-08
Migrate IapExample to 0.69.1 using template

andresesfm:
Migrated IapExample to latest RN. 
Removed apsl-react-native-button dependency as it is no longer compatible.


IapExample/App.js
```diff
@@ -19,7 +20,7 @@ import RNIap, {
 } from 'react-native-iap';
 import React, {Component} from 'react';
 
-import NativeButton from 'apsl-react-native-button';
+//import NativeButton from 'apsl-react-native-button';
```
hyochan, line 23: Looks like we can remove this

hyochan (approved):
Thanks for upgrading test project to recent RN 🙌

PR #1742, opened 2022-07-08
Master

nguyendinhnhan:


hyochan (changes requested):
@nguyendinhnhan Thanks so much for your work.

Could you kindly check the CI failing from the kotlin linter and request a review?

Also, I wish you can try to give more information on the title and the description.

android/src/amazon/java/com/dooboolab/RNIap/RNIapAmazonListener.kt
```diff
@@ -62,6 +68,27 @@ class RNIapAmazonListener(private val reactContext: ReactContext) : PurchasingLi
                     item.putString("productId", product.sku)
                     item.putString("price", priceNumber.toString())
                     item.putString("type", productTypeString)
+                    // Begin Add localizedPrice12
+                    val localizedPrice12 = if (priceNumber.toDouble().compareTo(0.00) > 0) {
+                        val price12 = priceNumber.toDouble().toBigDecimal().multiply(BigDecimal.valueOf(2))
+                        val currency = try {
+                            Currency.getInstance("VND")
```
andresesfm, line 75: Should this be a parameter? Or maybe it can be obtained from the payload? This makes it very specific to your locale

android/src/amazon/java/com/dooboolab/RNIap/RNIapAmazonListener.kt
```diff
@@ -19,6 +19,12 @@ import com.facebook.react.bridge.WritableNativeMap
 import com.facebook.react.modules.core.DeviceEventManagerModule.RCTDeviceEventEmitter
 import java.lang.NumberFormatException
 import java.util.ArrayList
+// Begin Add localizedPrice12
```
andresesfm, line 22: Linter complains about the order of these imports

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "8.2.1",
   "packageManager": "yarn@3.2.0",
   "description": "React Native In App Purchase Module.",
-  "main": "index.js",
+  "main": "index.ts",
```
andresesfm, line 6: This is only an issue when testing locally. For the release, this needs to be reverted to js

src/types/index.ts
```diff
@@ -45,6 +45,7 @@ export interface ProductCommon {
   price: string;
   currency: string;
   localizedPrice: string;
+  localizedPrice12: string;
```
andresesfm, line 48: What's the meaning of the '12'?

andresesfm:
Sorry but this doesn't seem generic-enough to be included in this library

PR #1744, opened 2022-07-13
Refresh Amazon products onResume as per documentation

andresesfm:
Amazon's documentation recommends fetching the products `onResume`


hyochan (approved):
Thanks for putting valid [reference](https://developer.amazon.com/docs/in-app-purchasing/iap-implement-iap.html#getpurchaseupdates-responses).

LGTM!

PR #1747, opened 2022-07-15
🛠️  : Fix  NativeModules.RNIapModule.getPackageName()

firemanx07:
this is required for deepLinkToAndroidSubscription(sku)

### DIFF
--- a/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
+++ b/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
@@ -609,7 +609,7 @@ class RNIapModule(
 
     @get:ReactMethod
     val packageName: String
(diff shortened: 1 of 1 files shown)


hyochan (commented):
@andresesfm Could you check if this is required?

andresesfm:
@firemanx07 Can you please tell us what you are trying to fix? It looks like you are only changing that method from Kotlin's property syntax to Java's old method call.

firemanx07:
@andresesfm   
deepLinkToSubscriptionsAndroid(sku: string) fails in finding the right  packageName of the application  => NatvieModules.RNIapModules.getPackageName() returns undefined .

andresesfm:
Closing in favor of: https://github.com/dooboolab/react-native-iap/pull/1748

PR #1749, opened 2022-07-18
First check for billing to be avaiable

andresesfm:
There is a crash on Android when attempting to test the app using an emulator/device without Play Services. The crash happens before the app initializes. This is because we were creating the billing client in the constructor instead of first checking it Google Play Services was available. This PR fixes it

### DIFF
--- a/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
+++ b/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
@@ -58,10 +58,7 @@ class RNIapModule(
             val nested = PromiseImpl(
                 {
                     if (it.isNotEmpty() && it[0] is Boolean && it[0] as Boolean) {
-                        val billingClient = billingClientCache
-                        if (billingClient != null) {
-                            callback(billingClient)
-                        } else {
+                        billingClientCache?.let(callback) ?: run {
                             promise.safeReject(DoobooUtils.E_NOT_PREPARED, "Unable to auto-initialize connection")
                         }
                     } else {
@@ -100,20 +97,21 @@ class RNIapModule(
             promise.safeResolve(true)
             return
         }
-        billingClientCache = builder.setListener(this).build()
+        billingClientCache = builder.setListener(this).build().also {
+            
+            it.startConnection(
+                object : BillingClientStateListener {
+                    override fun onBillingSetupFinished(billingResult: BillingResult) {
+                        if (!isValidResult(billingResult, promise)) return
 
-        billingClientCache?.startConnection(
-            object : BillingClientStateListener {
-                override fun onBillingSetupFinished(billingResult: BillingResult) {
-                    if (!isValidResult(billingResult, promise)) return
+                        promise.safeResolve(true)
+                    }
 
-                    promise.safeResolve(true)
-                }
-
-                override fun onBillingServiceDisconnected() {
-                    Log.i(TAG, "Billing service disconnected")
-                }
-            })
+                    override fun onBillingServiceDisconnected() {
+                        Log.i(TAG, "Billing service disconnected")
+                    }
+                })
+        }
     }
 
     @ReactMethod
--- a/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
+++ b/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
@@ -119,7 +119,7 @@ class RNIapModule(
     @ReactMethod
     fun endConnection(promise: Promise) {
         billingClientCache?.endConnection()
-
+        billingClientCache = null
         promise.safeResolve(true)
     }
 
--- a/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
+++ b/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
@@ -58,7 +58,10 @@ class RNIapModule(
             val nested = PromiseImpl(
                 {
                     if (it.isNotEmpty() && it[0] is Boolean && it[0] as Boolean) {
-                        billingClientCache?.let(callback) ?: run {
+                        val connectedBillingClient = billingClientCache
+                        if (connectedBillingClient?.isReady == true) {
+                            callback(connectedBillingClient)
+                        } else {
                             promise.safeReject(DoobooUtils.E_NOT_PREPARED, "Unable to auto-initialize connection")
                         }
                     } else {
--- a/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
+++ b/android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
@@ -40,21 +40,30 @@ class RNIapModule(
     ReactContextBaseJavaModule(reactContext),
     PurchasesUpdatedListener {
 
-    private var billingClient: BillingClient? = null
+    private var billingClientCache: BillingClient? = null
     private val skus: MutableMap<String, SkuDetails> = mutableMapOf()
     override fun getName(): String {
         return "RNIapModule"
     }
 
-    internal fun ensureConnection(promise: Promise, callback: () -> Unit) {
+    internal fun ensureConnection(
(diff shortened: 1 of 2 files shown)


android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -108,7 +109,7 @@ class RNIapModule(
 
     @ReactMethod
     fun endConnection(promise: Promise) {
-        billingClient.endConnection()
+        billingClient?.endConnection()
         promise.safeResolve(true)
```
hyochan, line 113: I think it'd be better to handle when `billingClient` is null like below.


```
if (billingClient) {
  billingClient.endConnection()
  promise.safeResolve(true)

  return
}

promise.safeReject(false) // billingClient is null
```

How do you think?

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -90,9 +83,17 @@ class RNIapModule(
...
 
+        if (billingClient?.isReady == true) {
+            Log.i(
+                TAG,
+                "Already initialized, you should only call initConnection() once when your app starts"
+            )
+            promise.safeResolve(true)
+            return
+        }
         billingClient = builder.setListener(this).build()
```
hyochan, line 94: Would there be any chance on `billingClient` isn't `null` but `isReady` is `false`?
If so, do you think the garbage collector here would make such a problem?

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -180,73 +181,72 @@ class RNIapModule(
         ) {
             val skuList = ArrayList<String>()
             for (i in 0 until skuArr.size()) {
-                val sku = skuArr.getString(i)
-                if (sku is String) {
+                if (skuArr.getType(i) == ReadableType.String) {
```
hyochan, line 184: Wouldn't `skuArr` always be the string array?

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -180,73 +181,72 @@ class RNIapModule(
...
+                    val sku = skuArr.getString(i)
                     skuList.add(sku)
                 }
             }
             val params = SkuDetailsParams.newBuilder()
             params.setSkusList(skuList).setType(type)
-            billingClient.querySkuDetailsAsync(
+            billingClient?.querySkuDetailsAsync(
```
hyochan, line 191: Should we reject if `billingClient` is null?

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -274,7 +274,7 @@ class RNIapModule(
             promise
         ) {
             val items = WritableNativeArray()
-            billingClient.queryPurchasesAsync(
+            billingClient?.queryPurchasesAsync(
```
hyochan, line 277: Same question here. Shouldn't we reject a case where `billingClient` is null?

hyochan (changes requested):
Thanks a lot, @andresesfm as usual.

I just left a few questions

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
andresesfm: Yes it should but ReadableArray holds Any type of objects. The old check for it to be string was redundant. In this way we guarantee that the elements are actually strings

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
andresesfm: I fixed it so it never gets into that case

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
andresesfm: same, now it's never null

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
andresesfm: I changed it so that it the instance gets disposed on `endConnection`

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
andresesfm: I've thought long and hard about this one. In the end, I think there are arguments for both ways of doing it. However, having to catch on a preventive `endConnection` call almost feels like punishing the developer that wanted to be extra careful. We can come back to it if needed

andresesfm:
@hyochan I think I covered all your comments. I think it's a lot cleaner now, let me know if any questions

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -90,9 +92,17 @@ class RNIapModule(
...
-        billingClient = builder.setListener(this).build()
+        if (billingClientCache?.isReady == true) {
+            Log.i(
+                TAG,
+                "Already initialized, you should only call initConnection() once when your app starts"
+            )
+            promise.safeResolve(true)
+            return
+        }
+        billingClientCache = builder.setListener(this).build()
```
hyochan, line 103: @andresesfm I think we can help the garbage collector here too.
```
billingClientCache = null
billingClientCache = builder.setListener(this).build()
```

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -90,9 +92,17 @@ class RNIapModule(
...
 
-        billingClient.startConnection(
+        billingClientCache?.startConnection(
```
hyochan, line 105: Do `billingClientCache` here to be optional?

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
andresesfm: That extra statement does not make a difference. The old reference will be garbage collected regardless 🙁

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
andresesfm: The reason is that smartcast doesn't know that the nullable instance is no longer null. However, I found a very nice way to resolve it by using the `also` operator. It looks cleaner now. please take a look

android/src/play/java/com/dooboolab/RNIap/RNIapModule.kt
```diff
@@ -100,20 +97,21 @@ class RNIapModule(
...
 
-                override fun onBillingServiceDisconnected() {
-                    Log.i(TAG, "Billing service disconnected")
-                }
-            })
+                    override fun onBillingServiceDisconnected() {
+                        Log.i(TAG, "Billing service disconnected")
+                    }
+                })
+        }
```
hyochan, line 117: @andresesfm This looks so clean and awesome! Such a great example to use `also` in Kotlin 👍

hyochan:
@andresesfm Do you have any idea why `./gradlew build test` is failing?

andresesfm:
> @andresesfm Do you have any idea why `./gradlew build test` is failing?

Yes, @hyochan I realized that the assignment would happen after the call to `also` so I rearranged the assignment to happen inside and it's now working. Good thing we have those tests 😅

hyochan:
> > @andresesfm Do you have any idea why `./gradlew build test` is failing?
> 
> Yes, @hyochan I realized that the assignment would happen after the call to `also` so I rearranged the assignment to happen inside and it's now working. Good thing we have those tests 😅

Fantastic!! You rocked 👊

hyochan (approved):

PR #1755, opened 2022-07-24
Fix for iOS IAP

rohitatcoffee:


rohitatcoffee:
Can someone help me to fix this check failure issue?

andresesfm:
@rohitatcoffee Thanks again. This new PR has so many changes that it is almost impossible to figure out what the actual fix is. Maybe if you isolate the change, we can convert it to swift? I can help you convert the core of your changes if you tell me what they are

PR #1756, opened 2022-07-25
Fixed function promotedProduct

Skarabei007:
promotedProduct always resolve null


andresesfm:
@Skarabei007 thank you for your PR. it looks correct. However, if you were getting null before, this won't change that

Skarabei007:
> @Skarabei007 thank you for your PR. it looks correct. However, if you were getting null before, this won't change that

This really solves my problem. In Swift I had a product but in js I had null. After this fix I get the product

PR #1759, opened 2022-07-25
Uses local ios files in Example app

andresesfm:
Having the Example use local ios files allows us to use IapExample to validate any ios changes

### DIFF
--- a/IapExample/react-native.config.js
+++ b/IapExample/react-native.config.js
@@ -1,0 +1,9 @@
+module.exports = {
+    dependencies: {
+      'react-native-iap': {
+        platforms: {
+          ios: null, // disable ios platform, so we can use local filess
+        },
+      },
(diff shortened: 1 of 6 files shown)


IapExample/react-native.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    dependencies: {
+      'react-native-iap': {
+        platforms: {
+          ios: null, // disable ios platform, so we can use local filess
+        },
+      },
+    },
+  };
```
hyochan, line 9: I'll just point out the reference for help.
https://github.com/react-native-community/cli/blob/main/docs/dependencies.md

Also the file looks like it has an extra tab. But we can handle that in the other PR.

hyochan (approved):
LGTM!

Issue #1697, opened 2022-05-18
getProducts and getSubscriptions return the same data


mtnt:
### Version of react-native-iap
8.0.8

### Version of react-native
0.68.1

### Platforms you faced the error (IOS or Android or both?)
iOS (15.4) (did not check android yet)

### Expected behavior
`getProducts` and `getSubscriptions` return separate lists of products and subscriptions

### Actual behavior
both methods return products and subscriptions according to productsIds passed into parameters

### Tested environment (Emulator? Real Device?)
Device (iPhone X)

### Steps to reproduce the behavior
```
await RNIap.initConnection()
this.purchaseUpdateSubscription = RNIap.purchaseUpdatedListener(() => {})
await RNIap.clearProductsIOS();

const products = await RNIap.getSubscriptions(['subscriptionId', 'concumableId']);
or 
const products = await RNIap.getProducts(['subscriptionId', 'concumableId']);
```

andresesfm:
We could filter them using https://developer.apple.com/documentation/storekit/skproduct/2936884-subscriptionperiod

andresesfm:
Fixed here: https://github.com/dooboolab/react-native-iap/pull/1760

PR #1760, opened 2022-07-25
Filtering subscriptions and iap products

andresesfm:
Fixes https://github.com/dooboolab/react-native-iap/issues/1697. Only including subscriptions/iap products in the result array

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -40,6 +40,8 @@ export enum InstallSourceAndroid {
 }
 
 export interface ProductCommon {
+  type: 'subs' | 'sub' | 'inapp' | 'iap';
+  productId: string;
   title: string;
   description: string;
   price: string;
@@ -112,12 +114,10 @@ export interface Discount {
 
 export interface Product extends ProductCommon {
   type: 'inapp' | 'iap';
-  productId: string;
 }
 
 export interface Subscription extends ProductCommon {
   type: 'subs' | 'sub';
-  productId: string;
 
   discounts?: Discount[];
 
--- a/src/iap.ts
+++ b/src/iap.ts
@@ -140,9 +140,10 @@ export const getProducts = (skus: string[]): Promise<Array<Product>> =>
   (
     Platform.select({
       ios: async () => {
-        const items = await getIosModule().getItems(skus);
-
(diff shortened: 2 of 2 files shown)


src/types/index.ts
```diff
@@ -40,6 +40,8 @@ export enum InstallSourceAndroid {
 }
 
 export interface ProductCommon {
+  type: 'subs' | 'sub' | 'inapp' | 'iap';
```
hyochan, line 43: We currently have 'iap' and 'sub' for iOS and
`inapp` and `subs' for [android which is the official type](https://developer.android.com/reference/com/android/billingclient/api/BillingClient.SkuType).

Looks like we can unify this.

Looks

hyochan (approved):
LGTM!

Dallas62:
This PR just beak subscriptions, but who cares...

image

Stop try to ONE-LINE everythings, you just filtered a Promise which leads to `undefined undefined is not a function [TypeError: undefined is not a function]`

hyochan:
> This PR just beak subscriptions, but who cares...
> 
> image
> 
> Stop try to ONE-LINE everythings, you just filtered a Promise which leads to `undefined undefined is not a function [TypeError: undefined is not a function]`

Oh my fault 😱  I should have review more carefully.

So the code should be
```
(await getIosModule()
          .getItems(skus))
          .filter((item: ProductCommon) => skus.includes(item.productId))
          .filter((item: ProductCommon) => item.type === 'iap');
```

@Dallas62 Could you kindly check if this works?

Dallas62:
Make it more readable, to avoid confusion:

```js
const items = await getIosModule().getItems(skus);

return items.filter((item: ProductCommon) => skus.includes(item.productId))
          .filter((item: ProductCommon) => item.type === 'iap');
```

(format/indent like you want)

hyochan:
> Make it more readable, to avoid confusion:
> 
> ```js
> const items = await getIosModule().getItems(skus);
> 
> return items.filter((item: ProductCommon) => skus.includes(item.productId))
>           .filter((item: ProductCommon) => item.type === 'iap');
> ```
> 
> (format/indent like you want)

Would you kindly submit a PR for this?

PR #1761, opened 2022-07-25
feat: catch error through context for init connection

jeremybarbet:


andresesfm:
Great, thank you  @jeremybarbet

Issue #1576, opened 2021-11-22
[iOS] AppAccountToken for App Store Server Notifications


AnthonyHaykal:
Hello,

iOS has a new JWSTransactionDecodedPayload which contains the following property:

appAccountToken: A UUID that associates the transaction with a user on your own service. If your app doesn’t provide an appAccountToken, this string is empty. For more information, see appAccountToken(_:).

Linked [here](https://developer.apple.com/documentation/appstoreservernotifications/jwstransactiondecodedpayload?changes=latest_major)

I would like to ask you if we can add the appAcountToken property when we purchase and get it from "App Store Server Notifications"

Thank you for all the hard work :).

ManetJun:
We have the similar requirement as this.
We also need to send the UUID as part of the product purchase option.
`Product.PurchaseOption.appAccountToken(UUID)`

Not sure if this is already in the roadmap and when could it probably be available?

Thanks.

Wanes-Tutunjian:
Hello

We are also facing this issue in our products, it would be a a lot easier if we can set a user identifier token for IOS purchases the way it is feasible for Android 

We were wondering if there is any update regarding this feature, and if it will be added anytime soon @hyochan 

Thanks for all your efforts :)

mklsk:
Likewise - would be awesome to have this capability added!

CC: @hyochan

sampsonjoliver:
Also +1'ing this.

As I understand this would be blocked by https://github.com/dooboolab/react-native-iap/issues/1403 ?

Wanes-Tutunjian:
> As I understand it, this would not be blocked by migrating to StoreKit v2.
> 
> In the original implementation of StoreKit there is a field in the `SKMutablePayment` object that allows you to tag a payment to a specific user with an arbitrary string.
> 
> If that string happens to be a `UUID`, then the receipt will include the `appAccountToken` field with that `UUID` value.
> 
> https://developer.apple.com/documentation/storekit/skmutablepayment/1506088-applicationusername
> 
> This would make `applicationUserName` field analogous to `appAccountToken`.

thanks for pointing this out... we added the applicationUsername parameter to the request subscription function and patched the package... works like a charm 👍🏿

He1nr1chK:
Hi @Wanes-Tutunjian, would you mind posting what you changed and how etc? Would love to do this myself. Thanks

jeffy2007:
[...]

PR #1763, opened 2022-07-27
feat(ios-purchase): add appAccountToken to requestPurchase on iOS

mmajk55:
According to the issue https://github.com/dooboolab/react-native-iap/issues/1576 there is a need for an extra param for iOS transactions.

### DIFF
--- a/ios/RNIapIos.m
+++ b/ios/RNIapIos.m
@@ -18,6 +18,7 @@ RCT_EXTERN_METHOD(getAvailableItems:
                   reject:(RCTPromiseRejectBlock)reject)
 RCT_EXTERN_METHOD(buyProduct:
                   (NSString*)sku
+                  appAccountToken:(NSString*)appAccountToken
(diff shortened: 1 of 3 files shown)


andresesfm:
Great job @mmajk55 . Quick Question, why call it `appAccountToken` and not `applicationUsername`?

jeremybarbet:
- This is missing update on the documentation
- That's a breaking change for folks using `requestPurchase(sku, false)`, since `appAccountToken` is now required
- We should probably pass to an object configuration so it's either to configure optional props
- We should mark this props as optional since it's only use in iOS side

mmajk55:
> ###

`appAccountToken` is recommended in the apple documentation https://developer.apple.com/documentation/appstoreserverapi/appaccounttoken

mmajk55:
@andresesfm True it should be called `applicationUsername `. I will change it 👍 . 
@jeremybarbet  This is a good idea to pass an object configuration and make this prop optional. I will create another PR with those changes.

PR #1770, opened 2022-07-28
Fixed promoted products when app doesn't start

Skarabei007:
When I add `RNIapIosQueue` to `AppDelegate` only sometimes promoted products are caught. After moving `SKPaymentQueue.default().add(self)` to `init` everything always works.


andresesfm:
Fantastic, thank you again @Skarabei007

KP-MobileTechie:
Fixed promoted products when app doesn't start not working in latest version

PR #1773, opened 2022-07-28
Fixed getSubscriptions and getProducts

Skarabei007:


andresesfm:
That works, thank you @Skarabei007

PR #1778, opened 2022-07-29
Publish beta on merge in beta branch

hyochan:
Manage the `next` release on the separate branch in `next` instead of publishing to the release tag with `npm`.
The downside of previous releases is that the user can't find the stable release of the version and we can't update the latest stable version when there has been a mistake.

Now, when `rebase merge` the `main` branch in `beta`, the CI will run and release the package to with `next` tag instead of the release.

- Related https://github.com/dooboolab/react-native-iap/discussions/1777

### DIFF
--- a/.github/workflows/publish-beta.yml
+++ b/.github/workflows/publish-beta.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: '16.x'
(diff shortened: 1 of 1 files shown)


hyochan:
Notsure why other CIs like `Ci Example iOS` are running altough their `path` is not correct 🤔

Issue #1780, opened 2022-08-01
Exception 'buyProduct:appAccountToken:andDangerouslyFinishTransactionAutomatically:applicationUsername:resolve:reject: is not a recognized Objective-C method


cdiddy77:
> Please use [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want get some help out of it. Please report issue if there is a bug.

### Version of react-native-iap

9.0.0-beta4

### Version of react-native

0.66.4

### Platforms you faced the error (IOS or Android or both?)

iOS

### Expected behavior

buyProduct works

### Actual behavior
```
Exception 'buyProduct:appAccountToken:andDangerouslyFinishTransactionAutomatically:applicationUsername:resolve:reject: is not a recognized Objective-C method.' was thrown while invoking buyProduct on target RNIapIos with params (
    "nft_purchase_1",
    0,
    fjSz2bdroEqLrTN62o0z16,
    11332,
    11333
)
callstack: (
    0   CoreFoundation                      0x000000018b0ad29c 5198FB57-5645-3B34-A49F-F32B52256CF3 + 627356
    1   libobjc.A.dylib                     0x00000001a3da7744 objc_exception_throw + 60
    2   Foundation                          0x000000018c93a4c8 AC6AD2F3-C7B1-38AA-8588-57036A663180 + 1254600
    3   app                                 0x00000001051b3bdc -[RCTModuleMethod processMethodSignature] + 836
    4   app                                 0x00000001051ba190 -[RCTModuleMethod invokeWithBridge:module:arguments:] + 140
    5   app                                 0x00000001051be364 _ZN8facebook5reactL11invokeInnerEP9RCTBridgeP13RCTModuleDatajRKN5folly7dynamicEiN12_GLOBAL__N_117SchedulingContextE + 1156
    6   app                                 0x00000001051bdd04 _ZZN8facebook5react15RCTNativeModule6invokeEjON5folly7dynamicEiENK3$_0clEv + 144
    7   app                                 0x00000001051bdc68 ___ZN8facebook5react15RCTNativeModule6invokeEjON5folly7dynamicEi_block_invoke + 28
    8   libdispatch.dylib                   0x000000018ad12e6c 355ACCF4-3917-3730-BC55-EF7003887ABE + 7788
    9   libdispatch.dylib                   0x000000018ad14a30 355ACCF4-3917-3730-BC55-EF7003887ABE + 14896
[...]

PR #1781, opened 2022-08-01
Fix mismatch between buyProduct signatues in Obj-C and swift

cdiddy77:
see https://github.com/dooboolab/react-native-iap/issues/1780


hyochan (changes requested):
Hello! Thanks for the fixes~!

However, I think this `PR` should be opened to `next` branch instead of `main`.
Since in `main`, `appAccountToken` is still used in iOS.

hyochan:
@cdiddy77 Please reopen the `PR` to `next` branch 🙏

PR #1786, opened 2022-08-01
refactor: new setup for the library

jeremybarbet:
This is the first PR of a few I would like to help with to refactor the structure of the library, provide better example, documentation, CI and clean up.

- I was looking at Amazon documentation and [they are sunsetting the SDK v2](https://developer.amazon.com/docs/in-app-purchasing/iap-migrate-iapv1-apps-to-iapv2.html) that react-native-iap is using.

	According to react-native-iap's documentation "the Amazon platform (Beta)." the support was still in beta. I am not sure how much Amazon is used as a payment provider compared to AppStore and Play, but it feels like to be a big effort to maintain the Amazon part.

	Since the SDK v2 will soon to removed, I would suggest to lean the core of react-native-iap and remove Amazon support. For users that would still depend on it they would need to stay on v8 (and v9?).

	It's just a suggestion, if someone is up for upgrading to the new SDK and maintaining, it's fine, but the smallest the scope of the library is, the easiest it is to maintain given the number of maintainers.

	It could also be added as some sort of plugin and developed on a separated repo (react-native-iap-amazon-plugin) for example.

---

Alright, so now, summary of this PR's changes:

## Structure
- Move to https://github.com/callstack/react-native-builder-bob for the library's setup. I created a new repo with the Kotlin/Swift template and moved over all the react-native-iap code. It allows a big clean up on dependencies/tooling that are not used anymore/deprecated and/or dead.
- Updated GA workflows
- Added swiftlint for iOS linting
- Added commitlint (and CI check) for better auto-generated changelog f.x.
	- commit message = `fix: something` when running `yarn release` it will automatically suggest `v8.0.0 -> v8.0.1`
	- commit message = `feat: something` -> `v8.0.0 -> v8.1.0`
	- commit message = `feat!: something` (! = breaking change) -> `v8.0.0 -> v9.0.0`
	- can also be controlled manually to define the wanted version 
- Added release-it for automatic release (github releases, changelog automation, npm publish), (ex: https://github.com/jeremybarbet/react-native-jsi-base-coder/releases)
- Removed flow/flowtype generation
- Removed jest, no test added, can be added back later
- Cleaned up non used devDeps
- Align as much as possible with react-native-builder-bob deps/devDeps
- Most of the LOC changes are on the yarn.lock files
- Replaced `@dooboo/eslint-config` with react-native-builder-bob's config. Is it an issue?
- Removed `dooboolab-welcome` that wasn't run, at least not with yarn. Good to be removed?
- Renamed `deploy-document` to `deploy-documentation`
- Renamed `IapExample` to `example`

## Sources
- Move the root index.ts file into src/index and merge it with other index.ts file
- Sources now built for commonjs/module
- Added eslint import-sort plugin
- Import `NativeEventEmitter` from `iap` index file rather than creating another reference into `withIAPContext`
- Added a `debugMessage` function in swift source to avoid printing the logs all the time

## Docs
- Run prettier on `.md` files as well
- Replace all `javascript` code blocks with `ts`

## Example
- Convert to typescript `App.js` to `App.tsx`, add missing types, but no others changes
- Does the bundle id has to be `com.iapexample`?

---

I was able to run the 3 examples:
- ios
- android/play
- android/amazon

If this PR looks good I would prefer to publish it under an alpha tag, just to make sure I can import it in my side project and check if it works fine

---

## Follow up PRs:
- [ ] Refactor types and move them to NativeModules directly 
- [ ] Refactor sources structure
- [ ] Clean up documentation, remove all deprecated mentions, add links to previous versions
- [ ] Expand the example app to use hooks/context and all the API methods
- [ ] Refactor publish GA workflows 
- [ ] `react-native-builder-bob` recommend to name native files as `RNIap` -> `Iap`. I could do that in a new PR if it's fine by you
- [ ] Rename the `RNIapModule` to `IapPlay` to follow the same logic (with the logic above)
- [ ] Rename `RNIapAmazonModule` to `IapAmazon`
- [ ] Rename `RNIapIos` to `IapIos`


andresesfm:
Hello @jeremybarbet. 
First of all, thank you for your contribution. I see a lot of positive changes here. Here are some notes to start a healthy discussion:
I see the following general areas that your PR is trying to cover:
Documentation update 👍 
Renaming RNIapAmazon: Should it be done separately? Should also change IapModule to IapAndroid?
IapExample -> example: I think it could be done on a separate PR. I personally have nothing against the change of name. Some notes though:  package name change would make it not compatible with current store registration I would assume. It uses an older version of rn: 68 vs the current 69. 
Removing Flow: I'm personally fine with this as we've moved to TS for a while. I'm not sure about everybody else though. 
React's new Architecture 👍 
There are some regressions. There are some files that revert code we recently merged 
Removing jest: We do have a test file `iap.test.ts`
Github Actions. 👍 
removing `monolinter`: I think we can still use it to keep example's depenencies in sync. 
Impact on v9.0.0 migration: As is with a single big PR is going to be difficult to merge going forward
Removing Amazon IAP. I wasn't aware of this being deprecated. I'd be willing to update/maintain the new SDK as it is something we use.
Min react-native. Why not go to 69?
Merge conflicts: need to be resolved. 

In general, I think we could break this down to make it more mangeable to review.

I hope these comments are the start of a constructive discussion. I think it's awesome what you've started here @jeremybarbet

jeremybarbet:
> I see the following general areas that your PR is trying to cover: Documentation update 👍
There are no changes on the documentation at the moment. Only prettier fixes. That could be split in another PR

> Renaming RNIapAmazon: Should it be done separately? Should also change IapModule to IapAndroid?
Very fair point, I will revert and do that on a follow up PR

> IapExample -> example: I think it could be done on a separate PR. I personally have nothing against the change of name.
Ditto

> Some notes though: package name change would make it not compatible with current store registration I would assume.
I wasn't sure about that, but that would make sense, will fix that

> It uses an older version of rn: 68 vs the current 69.
Good point, will update

> There are some regressions. There are some files that revert code we recently merged
The problem of doing a big PR like this, I probably missed some stuff. I will split in smaller PRs and review that

> Removing jest: We do have a test file `iap.test.ts`
I will add it back

> removing `monolinter`: I think we can still use it to keep example's depenencies in sync.
Same here

> Impact on v9.0.0 migration: As is with a single big PR is going to be difficult to merge going forward
I can work on the merge PR from 8 to 9 when needed 👍 

> Removing Amazon IAP. I wasn't aware of this being deprecated. I'd be willing to update/maintain the new SDK as it is something we use.
Sounds good to me

> Min react-native. Why not go to 69?
I just follow the react-native-builder-bob template here. I would probably say 0.69.x is still too early to be defined as minimal. Still a lot of libraries are not compatible with React 0.68.2. I would hold a couple of months more before moving to min 0.69.x. I will pick up the minimal versions defined before

Let me split this PR and revert some stuff around that should make this PR breath a bit. Thanks for going over it!

jeremybarbet:
@andresesfm Most of the changes have been merged in other PRs. Do you want me to split this more?

~98% of the LOC are coming from the 2 yarn.lock

Also, I am not a collaborator on npm so I can't create releases, but once this PR is merged, it would be nice to try the `yarn release` flow to make sure it create releases correctly on Github and generate the changelog file

jeremybarbet:
Ah yeah, for the example it's totally good for me. I already reverted the changes to use the 0.69.1 version on the example.

I will open a new PR for the example changes 👍

andresesfm:
@jeremybarbet The test to publish failed: https://github.com/dooboolab/react-native-iap/runs/7639944257?check_suite_focus=true#step:5:61

jeremybarbet:
@andresesfm Probably coming from the missing `registry-url: 'https://registry.npmjs.org'` Let me open a PR

docs/docs/usage_instructions/restoring_purchases.md
```diff
@@ -8,37 +8,41 @@ You can use `getAvailablePurchases()` to do what's commonly understood as “res
...
         case 'com.example.no_ads':
-          newState.ads = false
+          newState.ads = false;
           restoredTitles.push('No Ads');
-          break
+          break;
 
         case 'com.example.coins100':
-          await RNIap.finishTransaction(purchase.purchaseToken);
+          await RNIap.consumePurchaseAndroid(purchase.purchaseToken);
```
andresesfm, line 31: This should be `finishTransaction`

jeremybarbet:
Looks like we are getting there. Reverted a couple more files that can be done in others PRs and I'll be finally able to start on other exciting tasks now :D

android/build.gradle
```diff
@@ -1,91 +1,190 @@
...
-def DEFAULT_COMPILE_SDK_VERSION = 30
-def DEFAULT_BUILD_TOOLS_VERSION = "30.0.2"
-def DEFAULT_MIN_SDK_VERSION = 16
-def DEFAULT_TARGET_SDK_VERSION = 30
-def DEFAULT_PLAY_SERVICES_VERSION = "17.6.0"
+if (isNewArchitectureEnabled()) {
+  apply plugin: "com.facebook.react"
+}
+
+def getExtOrDefault(name) {
```
andresesfm, line 27: Really clean way of doing this

andresesfm:
> @andresesfm Probably coming from the missing `registry-url: 'https://registry.npmjs.org'` Let me open a PR

Yes, that was it. Thank you

docs/docs/usage_instructions/restoring_purchases.md
jeremybarbet: Thanks, not sure how this got change here

PR #1789, opened 2022-08-01
chore: configure yarn 2

jeremybarbet:
Extracted from https://github.com/dooboolab/react-native-iap/pull/1787

- Need a few other PRs to get merge first babel/metro config change, will keep it draft for now
- Should be based off `jeremybarbet:chore/prettier-config` but looks like it's not possible to do that with Github between fork and original repo..


jeremybarbet:
Will close that and merge within https://github.com/dooboolab/react-native-iap/pull/1787

PR #1797, opened 2022-08-02
refactor!: library source folder, types and methods

jeremybarbet:
Update:

I would appreciate a good review on that!

Summary of changes:

- Remove default export `import RNIap from 'react-native-iap'`. You now import by name directly e.g. `import { requestPurchase } from 'react-native-iap'`

- Restructure files:
  - context: contains the react context for hooks usage
  - hooks: contains the hook to work along with the context
  - internal: contains some utils used between files
  - modules: contains a file for each platform (ios/android/amazon), which their specific methods exposed (e.g. android — flushFailedPurchasesCachedAsPending, ios — presentCodeRedemptionSheetIOS). All the typescript interfaces are also defined per platform in there. Contains as well a `common.ts` that shares functions that are called by the 3 platforms (e.g. `requestPurchase`)
  - types: Typescript types. Most of the types are defined in `common.d.ts` because shared between all the modules
  - eventEmitter: React event listener to subscribe to.
  - purchaseError: Class for the Purchase error

Everything is exported at the root (apart from the internal folder)

---

So this PR splits:
- each modules in their own file with native typings directly
- remove some types non used and imported from external library
- get ride of all the jsdoc parameters to use directly in the javascript parameters to avoid naming the params twice each time
- get ride of the return types, it's coming from each modules directly

@andresesfm @hyochan 
- Do you have any objection on using external libraries for typings? I published theses two for apple/google types related to IAP `@jeremybarbet/apple-api-types`, `@jeremybarbet/google-api-types`. That's making some clean up in the types we were having.

- I would like to drop support for default export aka `RNIap.finishTransaction`, `RNIap.requestSubscription` and exports all theses methods through the destructured import `import { finishTransaction, requestSubscription } from 'react-native-iap'`. We don't really have a good process on how to handle breaking change at the moment, that can be done in another PR, just opening the conversation.

(@hyochan: is it fine I'm tagging you in here on kinda everything 😅? I don't want to spam you if you don't have time for it. Since I'm breaking/changing a lot of stuff, you might have opinion on it)


andresesfm:
@jeremybarbet I was thinking about this exact problem as I'm adding some other properties that are exclusive to Android. What would you think about typing objects like this:
```
type RequestPurchase{
commonProperty: type
android: {
   androidProp: type
}
ios: {
 otherProp: type
}
}
```
It would clean up a lot of our property names and maybe be less confusing that having them with the `never` type. Very curious about your thoughts on this one

jeremybarbet:
Yep, I like that, I think it would be an easy and clean way to treat platform-specific arguments.

`never` is really good when you **CAN'T** define two arguments at the same time, so if we need it somewhere I would still recommend using it. (cf: https://github.com/dooboolab/react-native-iap/pull/1802 we can use both android/ios arguments simultaneously so it didn't make sense to use `never` in there)

src/modules/index.ts
```diff
@@ -0,0 +1,19 @@
...
+    RNIapModule: AndroidModuleProps;
+    RNIapAmazonModule: AmazonModuleProps;
+  }
+}
+
+export * from './amazon';
+export * from './android';
+export * from './ios';
+
+export const NativeModule = isAndroid ? AndroidModule : IosModule;
```
andresesfm, line 19: Amazon missing here?

src/types/methods.ts
```diff
@@ -0,0 +1,169 @@
...
+  BuyProductIos,
+  UserDataAmazon,
+} from '../modules';
+
+/** Sku is a string that uniquely identifies a product/subscription */
+export type Sku = string;
+
+export enum ProductType {
+  /** Subscription */
+  'subs' = 'subs',
```
andresesfm, line 12: Should this be `subs = 'subs'`?

yarn.lock
```diff
@@ -10750,6 +10764,8 @@ __metadata:
   dependencies:
     "@babel/eslint-parser": 7.18.9
     "@firmnav/eslint-github-actions-formatter": 1.0.1
+    "@jeremybarbet/apple-api-types": 1.3.5
+    "@jeremybarbet/google-api-types": 1.3.5
```
andresesfm, line 10768: Cool, where are these used?

src/types/methods.ts
```diff
@@ -0,0 +1,169 @@
...
+  productId: string;
+  title: string;
+  description: string;
+  price: string;
+  currency: string;
+  localizedPrice: string;
+  countryCode?: string;
+}
+
+export interface ProductProduct extends ProductCommon {
```
andresesfm, line 84: Why `ProductProduct` what was the problem with `Product`?

src/types/amazon.ts
```diff
@@ -1,7 +1,7 @@
-/**
- * From: https://developer.amazon.com/es/docs/in-app-purchasing/iap-rvs-examples.html
- */
-export type ReceiptType = {
+import type {Sku} from './index';
```
andresesfm, line 1: Not sure I fully understand why use the `Sku` type alias, is there a benefit vs just using string?

andresesfm:
@jeremybarbet Thank you for such a fantastic job. I added a few questions.

I also wanted to discuss the plan of integrating this on main. Should we merge it on `next` once 19.0.0 lands? what are your thougs?

src/modules/index.ts
jeremybarbet: Amazon is checked [here](https://github.com/dooboolab/react-native-iap/pull/1797/files/beca8d361c6aedd4a19efc21dbecd4ceffe2a198#diff-32fc8165c1fa57159a39b1ef5a3c01c779827b25c9513f25360dc66cde072f8cR88). It's either iOS or Android as "device" modules, and Android can either be Play Store or Amazon IAP.

src/types/methods.ts
jeremybarbet: Sure, I will change that

yarn.lock
jeremybarbet: It removes most of the types for Apple and all of the ones for Google. It's used [here](https://github.com/dooboolab/react-native-iap/pull/1797/files/beca8d361c6aedd4a19efc21dbecd4ceffe2a198#diff-38940e5890f175d4fbdec6f5a090d634063b6f7c65b2690a2f1ee30a91a2e830R2-R6). (`src/methods.ts`, make sure to click "Load Diff", Github is hiding the most important part..)

src/types/methods.ts
jeremybarbet: Ah well! A big dilemma here.

We had [this type](https://github.com/dooboolab/react-native-iap/pull/1797/files/beca8d361c6aedd4a19efc21dbecd4ceffe2a198#diff-9463dd5d6cf72f3484ed3a3022651f3e8f1c0df3b03b553237a03c383bc867ccR169) before (`type Purchase = ProductPurchase | SubscriptionPurchase`). Instead of typing all the methods with `result: ProductPurchase | SubscriptionPurchase` we can then shortcut to `result: Purchase`. However, this is for a purchase response.

But for getting items (products or subscriptions) I had no idea on how to name the union type. Because we have products (consumable, non consumable) and subscriptions. which end up being `type Product = ProductProduct | SubscriptionProduct;` To keep consistency I wanted to find a better name, but couldn't think of anything else that would make sense

src/types/amazon.ts
jeremybarbet: Hum, theoretically, if the type of Sku were to change, it would be easy and be done in one place, instead of going over all the `sku: string`. In reality, that probably won't happen, but it just to give some extra type safety and define the type only in one place to make it clear.

jeremybarbet:
Theses changes are definitely breaking. The best move would be to merge and tag this in 9.x within the `next` branch.

andresesfm:
Partially integrated into 10.1.0

PR #1807, opened 2022-08-04
chore: new examples

jeremybarbet:
Alright alright, that's the new draft for the example app:
- Primarily showcasing the usage of context/hooks, which is probably the most common way to use the library nowadays
- Split by functionalities to make it easier to understand how to get/list/request purchase for each type
- kept the class example as well in another section

https://user-images.githubusercontent.com/937328/183248338-ce0b22cb-0a30-41bd-af27-d6ed725a6df4.mov

cc/ @andresesfm @hyochan
- Is there anything you have in mind that should be in the example app?
- I created a dummy example app on my appstore connect account so I can get the products showing, but to purchase through the sandbox server, the app likely need to be published? is there already a react-native-iap example published on the store?


jeremybarbet:
Also updated the design of the old `class` example
Screenshot 2022-08-06 at 12 50 52

PR #1809, opened 2022-08-05
[iOS] Fix issues related to pending transactions.

ifsnow:
There seems to be a problem while converting from objective-c to swift.

In a sandbox account with a lot of payment history, the following problems can be easily reproduced.

# Problems about `clearTransaction`
  * `countPendingTransaction` was not assigned ([Reference: 7.5.6 version](https://github.com/dooboolab/react-native-iap/blob/7.5.6/ios/RNIapIos.m#L264))
  * `resolve(nil)` is executed unconditionally ([Reference: 7.5.6 version](https://github.com/dooboolab/react-native-iap/blob/7.5.6/ios/RNIapIos.m#L273))

# Problem about observer
  * Not added to observer when calling `initConnection` again after calling `endConnection` ([Reference: 7.5.6 version](https://github.com/dooboolab/react-native-iap/blob/7.5.6/ios/RNIapIos.m#L115))

### DIFF
--- a/ios/RNIapIos.swift
+++ b/ios/RNIapIos.swift
@@ -44,6 +44,7 @@ class RNIapIos: RCTEventEmitter, SKRequestDelegate, SKPaymentTransactionObserver
   private var promotedProduct: SKProduct?
   private var productsRequest: SKProductsRequest?
   private var countPendingTransaction: Int = 0
+  private var hasTransactionObserver = false
 
   override init() {
     promisesByKey = [String: [RNIapIosPromise]]()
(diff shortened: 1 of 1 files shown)


andresesfm:
@ifsnow thank you for your contribution. This was recently changed here: https://github.com/dooboolab/react-native-iap/pull/1770

So I want to invite @Skarabei007 to share their opinion so we can avoid flip-flopping on adding the queue from the constructor to the init method.

- Constructor:
Fixes promoted items not being captured.

- initConnection:
-allows to end connection and reconnect again

- Third option??

andresesfm:
The other two changes look good @ifsnow (In case you want to separate them from the other issue. I think we could merge them right away

Skarabei007:
If you remove the add queue from initialization, promoted products will not be caught. I think we can check in initConnection that the connection has been created, and if it has been destroyed in endConnection, create a new connection.

ifsnow:
@andresesfm @Skarabei007 Thanks for feedbacks. I've modified it a bit considering what you said. 

Please let me know if we need further thought or need to split this PR.

ifsnow:
@andresesfm Version 8.6.5 was not published in [NPM](https://www.npmjs.com/package/react-native-iap). Please check.

andresesfm:
That is correct @ifsnow. This is the reason: https://github.com/dooboolab/react-native-iap/actions/runs/2813166039. I will fix it in the morning.
@jeremybarbet looks like the build used to publish is different from the one used on PRs.

jeremybarbet:
Ah yeah, I know why. I'm opening a PR to fix that

hyochan:
I've looked into this and found that this is due to the `IapExample` packages not being installed before publishing.
It looks like we should separate the role here since the example project should not bother the publishing stages.

PR #1810, opened 2022-08-05
feat(ios): pass clearTransactionIos function to context

mmajk55:
Pass `clearTransactionIos` to Iap context

### DIFF
--- a/src/hooks/useIAP.ts
+++ b/src/hooks/useIAP.ts
@@ -1,6 +1,7 @@
 import {useCallback} from 'react';
 
 import {
+  clearTransactionIOS as iapClearTransactionIOS,
   finishTransaction as iapFinishTransaction,
   getAvailablePurchases as iapGetAvailablePurchases,
   getProducts as iapGetProducts,
@@ -28,6 +29,7 @@ type IAP_STATUS = {
     isConsumable?: boolean,
     developerPayloadAndroid?: string,
   ) => Promise<string | void>;
+  iapClearTransactionIOS: () => Promise<void>;
   getAvailablePurchases: () => Promise<void>;
   getPurchaseHistories: () => Promise<void>;
   getProducts: (skus: string[]) => Promise<void>;
@@ -120,6 +122,7 @@ export function useIAP(): IAP_STATUS {
     currentPurchaseError,
     initConnectionError,
     finishTransaction,
+    iapClearTransactionIOS,
     getProducts,
     getSubscriptions,
     getAvailablePurchases,


src/hooks/withIAPContext.tsx
```diff
@@ -6,6 +6,7 @@ import {
   promotedProductListener,
   purchaseErrorListener,
   purchaseUpdatedListener,
+  clearTransactionIOS,
```
jeremybarbet, line 9: You don't need to edit the withIapContext file, the function can be imported straight away in the useIAP hook similar as `finishTransaction` and `finishTransaction as iapFinishTransaction`

src/hooks/useIAP.ts
```diff
@@ -34,6 +34,7 @@ type IAP_STATUS = {
   getSubscriptions: (skus: string[]) => Promise<void>;
   requestPurchase: typeof iapRequestPurchase;
   requestSubscription: typeof iapRequestSubscription;
+  clearTransactionIOS: () => Promise<void>
```
jeremybarbet, line 37: Can you move this under the `finishTransaction` method to group by "logic" methods, and you can import it straight away from this file as 

```ts
import {clearTransactionIOS as iapClearTransactionIOS} from './iap'
```

jeremybarbet (changes requested):

jeremybarbet:
A few comments, also, once done, make sure to run `yarn format` to apply prettier to your file and/or `yarn lint` to run tsc/prettier/eslint to your changes

src/hooks/withIAPContext.tsx
mmajk55: true 👍

hyochan (approved):
LGTM!

However, I want to point out that having this function exposed in the hook is a bit redundant since this can be called using `clearTransactionIOS` function in `react-native-iap` itself.

The functions exposed in the hook are intended as they needed the hook's lifecycle, such as having products updated declaratively when fetching them.

jeremybarbet:
@hyochan I agree on that, I was gonna make this comment as well. Not sure if it's worth it exposing through the hook as well

jeremybarbet:
@hyochan We already expose `requestPurchase` and `requestSubscription` which are not doing anything through the context. To keep consistency I would probably remove everything from the hook that doesn't have any interaction with the context

hyochan:
> @hyochan We already expose `requestPurchase` and `requestSubscription` which are not doing anything through the context. To keep consistency I would probably remove everything from the hook that doesn't have any interaction with the context

Thanks for clarifying this. I think what's most important is the documentation since there will be more contributors coming to update similar changes. I can work on the documentation later.

mmajk55:
I didn't know that this function can be called directly from RIap. Thanks for clarifying. So I think that we can close this PR

jeremybarbet:
> > @hyochan We already expose `requestPurchase` and `requestSubscription` which are not doing anything through the context. To keep consistency I would probably remove everything from the hook that doesn't have any interaction with the context
> 
> Thanks for clarifying this. I think what's most important is the documentation since there will be more contributors coming to update similar changes. I can work on the documentation later.

I started reworking the documentation on this [PR](https://github.com/dooboolab/react-native-iap/pull/1815) Feel free to jump in and add to it! (I will revert my "wip" commit, so it has a better commit message)

jeremybarbet:
[I removed ](https://github.com/dooboolab/react-native-iap/pull/1797/commits/770858645a4fd465a63e5b057b0e1a9d56bf321f)`requestPurchase` and `requestSubscription` from the hook to keep consistency.

PR #1813, opened 2022-08-06
fix: make all types optional for requestPurchase

jeremybarbet:
All theses types are being optional on native side and have either default value or being undefined on TS side

Also, question @andresesfm, we have `requestSubscription` to buy a subscription, wouldn't it make more sense to have `requestProduct` for a product, rather than `requestPurchase`? (could be done on another PR, just a curiosity question)

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -150,14 +150,14 @@ export interface Subscription extends ProductCommon {
 
 export interface RequestPurchase {
   sku: Sku;
-  andDangerouslyFinishTransactionAutomaticallyIOS: boolean;
+  andDangerouslyFinishTransactionAutomaticallyIOS?: boolean;
   applicationUsername?: string;
   obfuscatedAccountIdAndroid?: string;
   obfuscatedProfileIdAndroid?: string;
-  selectedOfferIndex: number;
+  selectedOfferIndex?: number;
 }
 
 export interface RequestSubscription extends RequestPurchase {
   purchaseTokenAndroid?: string;
-  prorationModeAndroid: ProrationModesAndroid;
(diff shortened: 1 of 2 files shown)


hyochan (approved):
All looks good to me.
Thanks for the hard work these days 🙇🏻

jeremybarbet:
Do you mind merging? I don't have rights for merging PRs

hyochan:
> Do you mind merging? I don't have rights for merging PRs

I just wanted to wait for @andresesfm since he is driving the repo currently. Do you think this should be urgently merged? @jeremybarbet

src/types/index.ts
```diff
@@ -150,14 +150,14 @@ export interface Subscription extends ProductCommon {
 
 export interface RequestPurchase {
   sku: Sku;
-  andDangerouslyFinishTransactionAutomaticallyIOS: boolean;
+  andDangerouslyFinishTransactionAutomaticallyIOS?: boolean;
```
hyochan, line 153: Looks like this is pretty urgent

hyochan:
I have managed to release current update to `8.6.4`.

Please take this commitment to `next` branch if necessary!

src/types/index.ts
jeremybarbet: yeah, it must be throwing a lot of eslint errors on folks' projects since it has been introduced in an earlier version and these props were optional before

hyochan:
I also added @jeremybarbet to `maintainer` since he looks like a driver now 🏃

jeremybarbet:
Thank you @hyochan I appreciate that!

PR #1815, opened 2022-08-06
docs: update documentation to match latest API

jeremybarbet:


andresesfm:
@jeremybarbet what are your thoughts on using https://typedoc.org/ to generate the docs vs separately maintaining it?

jeremybarbet:
I looked into it when starting this task and try [this plugin](https://github.com/tgreyuk/typedoc-plugin-markdown/tree/master/packages/docusaurus-plugin-typedoc) for docusaurus but I wasn't really convinced.

An example I have in mind, is for MST, I love MST, but in my opinion [the typedoc documentation](https://mobx-state-tree.js.org/API/) is just unreadable and non-human example friendly. I rather dig into the code to find what I need.

Ideally if this jsdoc:
Screenshot 2022-08-15 at 08 13 46

could generate a single markdown page like this, that would be the dream
Screenshot 2022-08-15 at 08 14 28

But I didn't find anything like that, at least not yet. I was also thinking to write our own code parser to create the documentation, but probably a lot of work.

So in short, I'm not against it, but I didn't find the library that answer this problematic

andresesfm:
@jeremybarbet any chance we could get this merged?

andresesfm:
Closing in favor or version resolving conflicts: https://github.com/dooboolab/react-native-iap/pull/1927

PR #1817, opened 2022-08-08
fix: install deps for example

jeremybarbet:


hyochan (commented):
@jeremybarbet I've just found the issue on the package not being published because the `example` build failed.

IMO, since the breakage of `example` is not related to publication, I think we can exclude this in publish `CI` and test this in other `ci`.

How does this sound to you?

jeremybarbet:
Yep, that make sense to me, it's a fair point. I can rework that a bit later on today

hyochan:
@jeremybarbet I am not sure if you'd like the approach here #1820 #1819.

I find this is the easiest since we are checking lint with various steps.
Screen Shot 2022-08-08 at 6 16 11 PM

To handle that, we need to separate the `eslintignore` environment for `development` and `production` which would solve the problem bit more complicated.

jeremybarbet:
@hyochan The error appears because I removed the `eslint` file from the example app [here](https://github.com/dooboolab/react-native-iap/pull/1807/files). I thought it was gonna be easier to just have one to eslint everything. I didn't think of the case where we run eslint without the example.

I would say, it's easier to add back the `eslint` for the example app. I'm not sure if we do an `extends` on the eslint config it will throw the same error or not, or just copy the same config as we have at the root, and add a new script `lint:eslint:example` and `lint:ci:example` that we would run only on example CI worflows would make it easier. 

Alternatively, we could create all these scripts within the example/package.json so we can run them individually avoiding confusion from the root package

hyochan:
@jeremybarbet IMO, I think we can just move on until we think this should be migrated. It'd be more productive to focus on the major things then come back later :)

PR #1801, opened 2022-08-03
fix: limit promotedProductListener to iOS

jgillick:
This attempts to fix issue: https://github.com/dooboolab/react-native-iap/issues/1800

I'm submitting this as a recommendation, I haven't had a chance to test it.

Issue #1800, opened 2022-08-03
Android: E_IAP_NOT_AVAILABLE due to promotedProductListener call


jgillick:
Version: 8.6.0

Using `withIAPContext` on Android throws the `E_IAP_NOT_AVAILABLE` because it calls the iOS-only method `promotedProductListener`.

The source of the bug is this line:
https://github.com/dooboolab/react-native-iap/blob/d5adf25f2220a2a4a334aa10fb4bd378f039c2fb/src/hooks/withIAPContext.tsx#L138-L145

From PR: https://github.com/dooboolab/react-native-iap/pull/1786

andresesfm:
Fix available in latest release

PR #1818, opened 2022-08-08
fix: included #1801 on next branch

SimonDamberg:
I am experiencing #1800 on latest beta release, since changes in #1801 is missing. This adds the same changes to the next branch.


SimonDamberg:
Nevermind, the changes seems to already be on the next branch, but missing from the releases.

PR #1829, opened 2022-08-09
chore: sync with main

jeremybarbet:
- Merging main into next
- After that I will change the base off `next` for this PR so it can be merged inside `next` instead of `main`


src/types/index.ts
```diff
@@ -171,23 +172,21 @@ export interface SubscriptionIOS extends ProductCommon {
 
 export type Subscription = SubscriptionAndroid & SubscriptionIOS;
 
-export interface RequestPurchaseBaseAndroid {
+export interface RequestPurchaseAndroid {
+  skus?: Sku[];
```
andresesfm, line 176: `Sku[]` is only for one-time purchases. Subscriptions require `SubscriptionOffer[]` instead

.github/workflows/ci.yml
```diff
@@ -67,5 +67,8 @@ jobs:
       - name: Build typescript
         run: yarn lint:ci
 
+      - name: Check commit message is valid
```
jeremybarbet, line 70: Was that removed on purpose?

The purpose of this check isn't fully completed, but it's intended to work along https://github.com/release-it/conventional-changelog#infile to generate/public automatically to npm and create the release changelog. Similar to [this](https://github.com/react-navigation/react-navigation/releases/tag/%40react-navigation%2Fstack%406.0.0-next.11)

andresesfm, line 70: I was going to propose we remove this as it becomes hard when merging across branches. It usually fails the build. Could it be a warning? Please note that we get some random PRs from people and I don't want that this strict rule turns them away from merging a single PR

jeremybarbet: Alright, I will remove that. We can revisit later on if needed 👍

src/types/index.ts
jeremybarbet: Thanks for spotting that, I will revert it

PR #1830, opened 2022-08-09
fix: update CI workflow

jeremybarbet:


.github/workflows/ci.yml
```diff
@@ -34,8 +36,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
```
andresesfm, line 38: Why remove this? doesn't it make the checkout faster?

jeremybarbet: No, unfortunately it's the opposite. `0` fetch all the history and tags (https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches)

Issue #1831, opened 2022-08-10
Handle IOS promocode and implement promoted subscriptions - RNIap/RNIap-Swift.h' file not found


atech-basemalnabulsi:
> Please use the [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want to get some help. Please use issues to report bugs.

**Description**

I am trying to use Promocode and promotion service form APPLE since it's required now to use it and not using our own promocode implementation.
While trying to define the Queue we faced import issue as "'RNIap/RNIap-Swift.h' file not found"
Based on RNIap documentation we should import it to define purchase queue, can you help to find the missing part.

Documentation URL is:
https://react-native-iap.dooboolab.com/docs/faq/#objc-version

When I comment the import I get error on 
[[SKPaymentQueue defaultQueue] addTransactionObserver:RNIapQueue.shared];
as "Property 'shared' not found on object of type 'RNIapQueue'"

**Expected Behavior**

We expect to accept the package or has a file called RNIap-Swift.h to import 
[[SKPaymentQueue defaultQueue] addTransactionObserver:RNIapQueue.shared];=

**Screenshots**

**Environment:**

- react-native-iap: ^8.6.5
- react-native: 0.63.4
- Platforms (iOS, Android, emulator, simulator, device): ios device

**To Reproduce**

1. Go to the URL https://react-native-iap.dooboolab.com/docs/faq/#objc-version
2. Continue with the steps
3. You will get the error in the second import step

---

[Optional] **Additional Context**

<img width="726" alt="Screen Shot 2022-08-10 at 5 07 56 AM" src="https://user-images.githubusercontent.com/109934170/183794742-7538fe47-60aa-41f7-8615-56544ce3040e.png">
<img width="721" alt="Screen Shot 2022-08-10 at 5 06 19 AM" src="https://user-images.githubusercontent.com/109934170/183794763-e993d5b4-4c3b-480a-8c39-211a0b3b3b90.png">

andresesfm:
`RNIapQueue` has been recently removed as it is no longer necessary. Documentation needs to be updated

andresesfm:
https://github.com/dooboolab/react-native-iap/pull/1836

PR #1836, opened 2022-08-10
doc: Removed manual native step for promoted products

andresesfm:
Fixes: https://github.com/dooboolab/react-native-iap/issues/1831


andresesfm:
@ifsnow, @Skarabei007 Please let me know if I'm correct about removing this part of the documentation

Skarabei007:
Yes, that is right

PR #1840, opened 2022-08-10
fix: bring back TVOS. It was removed in a recent refactoring

andresesfm:
Originally added here: https://github.com/dooboolab/react-native-iap/pull/869


jeremybarbet:
Nice, good catch!

jeremybarbet (approved):

PR #1841, opened 2022-08-10
feat: new android billing client v5

jeremybarbet:
Co-authored-by: Andres Aguilar <andresesfm@gmail.com>

---

I created a new branch base off `main`. Then cherry pick your commits from `next` branch and created this PR. That's 10 files changed.

When trying to merge next into main, it gives a 54 files changed.

I looked a bit and it could be how the base of each branches got "generated" (cf: https://stackoverflow.com/a/35028640/1913044). Honestly I don't really want to understand too much why ahah

Do you think there is a way to get this diff as a the clean state for the `next` branch? 

I would like to base my 2 open PRs off `next` but the current git history makes it very hard.

I see that `next` is protected:
- we could unprotect it, merge force
- use a new branch as "next"
- maybe something else

Let me know what you think 😊

### DIFF
--- a/.github/workflows/publish-next-package.yml
+++ b/.github/workflows/publish-next-package.yml
@@ -1,11 +1,11 @@
-name: Publish / Package
+name: Publish / Next Package
 
 on:
   release:
-    types: [created]
+    types: [prereleased]
 
 jobs:
-  build:
+  publish-npm-registry:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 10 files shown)


.github/workflows/publish-next-package.yml
```diff
@@ -1,11 +1,11 @@
-name: Publish / Package
```
andresesfm, line 1: I think these are mean to be two separate files, they shouldn't be merged into one

jeremybarbet:
Well obviously I must have missed some cherry pick.

I will try again to sync next with main, so we get a proper diff, before getting into creating this PR. Alternatively I can wait next to be merge and I will continue my work afterward

PR #1873, opened 2022-08-17
refactor!: make all methods take an object of params

jeremybarbet:
I flatten the interfaces within the methods to simplify next step of rework. I will start moving (in another PR) all modules specific params as we talked some time ago. It will be as follow:

```ts
requestPurchase({
  skus: [''], // common
  ios: {
    ...
  },
  android: {
    ..
  }
})
```

## Breaking changes

- All the exposed methods are now taking an object of parameters, rather than named params. e.g. `requestSubscription('com.example.sub', false)` -> `requestSubscription({ sku: 'com.example.sub', andDangerouslyFinishTransactionAutomaticallyIOS: false })`
- `requestPurchase` now takes an array of `Sku` -> `requestPurchase({ skus: ['com.example.sub'] })`. For iOS, only one Sku can be defined, for Android, it can be multiple
- Remove the `InAppPurchase` type, which was already defined as `ProductPurchase`

### DIFF
--- a/IapExample/src/screens/Products.tsx
+++ b/IapExample/src/screens/Products.tsx
@@ -26,7 +26,7 @@ export const Products = () => {
 
   const handleGetProducts = async () => {
     try {
-      await getProducts(constants.productSkus);
+      await getProducts({skus: constants.productSkus});
     } catch (error) {
       errorLog({message: 'handleGetProducts', error});
     }
@@ -34,7 +34,7 @@ export const Products = () => {
 
   const handleBuyProduct = async (sku: Sku) => {
     try {
-      await requestPurchase({sku});
+      await requestPurchase({skus: [sku]});
     } catch (error) {
       if (error instanceof PurchaseError) {
         errorLog({message: `[${error.code}]: ${error.message}`, error});
@@ -48,7 +48,11 @@ export const Products = () => {
     const checkCurrentPurchase = async () => {
       try {
         if (currentPurchase?.transactionReceipt) {
-          await finishTransaction(currentPurchase, true);
+          await finishTransaction({
+            purchase: currentPurchase,
+            isConsumable: true,
+          });
+
           setSuccess(true);
         }
       } catch (error) {
--- a/IapExample/src/screens/Products.tsx
+++ b/IapExample/src/screens/Products.tsx
@@ -34,7 +34,7 @@ export const Products = () => {
 
   const handleBuyProduct = async (sku: Sku) => {
     try {
-      await requestPurchase({skus: [sku]});
+      await requestPurchase({sku});
     } catch (error) {
       if (error instanceof PurchaseError) {
         errorLog({message: `[${error.code}]: ${error.message}`, error});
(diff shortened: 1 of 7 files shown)


andresesfm:
Hello @jeremybarbet, I like the change in general, but have a comment about the "common" part. I think we should not have a "common" fields section. It makes it unnecessarily complex and to be quite honest, I haven't seen many people use the same skus for android and iOS.  Having it as an array for iOS also makes it confusing as it may lead to confusion or having to do extra validation.

On a separate (but related to this) topic, how about amazon? I was debating between using `{ios:{...}, android:{play:{...}, amazon:{...}}}` or `{ios:{...}, play:{...},amazon:{...}}`. I.e nested vs flat

What are your thoughts?

jeremybarbet:
So for example, for this method:
Screenshot 2022-08-18 at 07 48 13

Would you also move to `getProducts({ ios: { skus: [] }, android/play: { skus: [] } })`? I would personally keep it the same way as before `getProducts({ skus: [] })` 

I agree it's a bit confusing for the iOS `requestPurchase`. I just thought it could be nice to have consistency on the usage of certain props, but both `sku` (iOS) and `skus` (android/play/amazon) would then be moved in their specific params object, for this method.

Oh yeah that's a good point as well. I like the `play`/`amazon`, and to go a step further, maybe rename ios to `appstore` to follow the logic of what they are defining.

andresesfm:
> getProducts({ ios: { skus: [] }, android/play: { skus: [] } })
Not like this ^^ but I'd be like: `getProducts({ ios: { sku: string }, android/play: { skus: [] }, amazon:{sku} })`
Or even `getProducts({ ios: IosRequest, android/play:PlayRequest, amazon:AmazonRequest } | SingleSkuRequest)`. Where `SingleSkuRequest = Sku` 😮, What do you think about this one?

jeremybarbet:
iOS takes an array of Sku for the [`getItems`](https://github.com/dooboolab/react-native-iap/blob/main/ios/RNIapIos.swift#L166) method (which is getProducts on JS side) hence my question to shorten a bit everything.

Can you give examples for each platform, what would it be for this call? `getProducts({ ios: IosRequest, android/play:PlayRequest, amazon:AmazonRequest } | SingleSkuRequest)`

If we allow more than one type of params, f.x. object and string (= Sku) it will require even more validation on the JS side to be able to check with params are passed or not.

I will revert the change I did regarding sku/skus so we can get this PR merge and we can continue the discussion for the new params format on the next PR 🙌

andresesfm (approved):

IapExample/src/screens/Products.tsx
```diff
@@ -38,7 +38,7 @@ export const Products = () => {
 
   const handleBuyProduct = async (sku: Sku) => {
     try {
-      await requestPurchase({sku});
+      await requestPurchase({skus: [sku]});
```
hyochan, line 41: If `iOS` does not accept the array, I think it'd be good better to alert user when array is given on `iOS`.

Maybe we can throw an error when the array has passed on `iOS` only?

andresesfm (approved):

PR #1861, opened 2022-08-16
add missing proration mode

mmajk55:
ProrationModesAndroid enum was missing IMMEDIATE_AND_CHARGE_FULL_PRICE mode

PR #1877, opened 2022-08-22
ios storekit2

andresesfm:
- set min ios version to 15
- consolidated buy methods
- removed checks for older versions of ios
- cleared most errors
- swiftlint
- ci: publishing packages dedicated to branches (#1859)
- add missing proration mode (#1861)
- 9.0.2
- ci: remove ref check on publishing main package
- continue migration, purchases
- return promises to class
- moved utils to ios
- Check for initalized listener before onResume happens (#1867)
- 9.0.3
- clean up promises and error codes
- Separate Amazon purchase args (#1870)
- 9.0.4
- serialized Transactions
- removed remaining old methods, added serialization
- default to Xcode 4 spaces
- Split files
- Added more transaction methods
- removed global autofinish
- fix lint on doc


andresesfm:
Merged

PR #1896, opened 2022-08-29
docs: Fixed getProducts signature

Abdelalim-dev:
In the source code, we have an object with one attribute: `skus`

```
export declare const getProducts: ({ skus, }: {
    skus: string[];
}) => Promise<Array<Product>>;
```


andresesfm:
Thank you @Abdelalim-dev

PR #1926, opened 2022-09-06
Fix subscriptionOffers promise rejected

likesebiii:
The return was invalid.

Promise rejected was called without being returned, thus it didn't trigger the catch and it was handled inside finally statement.


andresesfm (approved):

andresesfm:
Thank you @likesebiii

PR #1927, opened 2022-09-06
Docs/refactor

andresesfm:
Integrating Jeremy's work onto main


jeremybarbet:
It wasn't all finished, but most of it was there. Theses pages ([one](https://react-native-iap.dooboolab.com/docs/old-to-remove-available-purchase), [two](https://react-native-iap.dooboolab.com/docs/old-to-remove-product)) should be rewritten and moved to [product](https://react-native-iap.dooboolab.com/docs/api-reference/interfaces/product) and [subscription](https://react-native-iap.dooboolab.com/docs/api-reference/interfaces/subscription)

I'm stuck on big project deadline at the moment, so I will get back to it once I freed up some time. I would like to generate all the functions documentation using [Tsdoc](https://github.com/microsoft/tsdoc/tree/main/api-demo). They have some node module that we could leverage to generate all the pages based on the code directly like we discuss some time ago.

Cheers!

PR #1974, opened 2022-09-20
fix: modified the type of timestamp in the paymentDiscount from number to string

khj3132:
- modified the type of `timestamp` in the `paymentDiscount` from number to string.

- I modified this way because the timestamp property, which is a number type, causes a type casting error in the iOS environment.

- If it is better to fix the `buyProductWithOffer` in the `RNIapIos.swift` file than to fix the type like this, you can reject this PR.


andresesfm:
Closing in favor of: https://github.com/dooboolab/react-native-iap/pull/1980. Nothing wrong with this solution, thank you @khj3132. The Other PR handles the conversion internally, so we can keep the type checking

andresesfm:
Released as part of `10.1.3`

PR #1975, opened 2022-09-20
fix: added the offerTags property to the SubscriptionAndroid

khj3132:
- added the `offerTags` property to the `SubscriptionAndroid`

- There is no the `offerTags` property in the `SubscriptionAndroid` interface, so a lint error occurs when using the `offerTags`. So I added the `offerTags` property like this, but if it's unnecessary, you can reject it.


andresesfm:
Thank you @khj3132

PR #2038, opened 2022-10-11
Lint and run tests as part of yarn prepare to improve safety

hsource:
## Motivation

While trying to work on this repository, I realized that the linting and tests were broken. To help ensure the tests and linting isn't broken by any new changes, I think it'd be great to run the linter and tests as part of the `prepare` release process.

## Fix

1. Change `yarn prepare` to also run `yarn lint && yarn test`
   - Create a separate `yarn build` command so we can build while skipping the lint/test process
2. Fix the test by moving the mock to the outermost scope
3. Fix the linting by adding excludes to the typechecker

## Testing

Ran `yarn prepare` and verified that it runs successfully.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,3 @@
-__tests__
 node_modules
 *.js
 *.d.ts
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,7 +2,7 @@ module.exports = {
   clearMocks: true,
   coverageDirectory: 'coverage',
   moduleDirectories: ['node_modules', 'src'],
-  modulePathIgnorePatterns: ['IapExample'],
+  modulePathIgnorePatterns: ['IapExample', 'lib'],
   preset: 'react-native',
   setupFiles: ['<rootDir>/test/mocks/react-native-modules.js'],
   setupFilesAfterEnv: ['@testing-library/jest-native/extend-expect'],
--- a/package.json
+++ b/package.json
@@ -55,7 +55,8 @@
     "!**/__mocks__"
   ],
   "scripts": {
-    "prepare": "bob build",
+    "prepare": "yarn lint && yarn test && yarn build",
+    "build": "bob build",
     "release": "release-it",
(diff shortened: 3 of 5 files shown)


andresesfm:
Hello @hsource thank you for the contribution. Can you please describe how this change "improves safety". Also if you move the mock outside of `describe` how would you ad ios tests?

hsource:
@andresesfm 

> Hello @hsource thank you for the contribution. Can you please describe how this change "improves safety".

Thanks for the questions! What I meant by this was that the typechecking (`yarn lint:tsc`) and tests (`yarn test`) seemed to have errors that weren't caught by CI.

Thinking about this more after your comment, I think we can revert this change and change `prepare` back to just run `bob build` if we fix `yarn lint:tsc` in CI and add `yarn test` to the CI pipeline.

**Summary - I think instead of doing this, we may want to just add `yarn test` and make sure `yarn lint:tsc` is working on Github Actions for the CI pipelines**

Here's some more details:

When I ran `yarn lint:tsc` on the commit [11470abc](https://github.com/dooboolab/react-native-iap/commit/11470abc948e3072a04c54d276b9d18da4a1c206), it found 7 errors, only in files in the IapExample. I fixed these errors by excluding the IapExample folder. I think these errors weren't caught by CI because the ["Run typescript"](https://github.com/dooboolab/react-native-iap/actions/runs/3220920047/jobs/5268222753#step:7:7) job isn't completing because of a reviewdog error ("This GitHub token doesn't have write permission of Review API").

When I ran `yarn test` on 11470abc, I also got an error. To make sure we don't break the tests again after I fixed it in this PR, I think we should add `yarn test` to the CI pipeline.

> Also if you move the mock outside of `describe` how would you ad ios tests?

Right now, there aren't any iOS tests, so it's not a problem. `jest.mock` only seemed to work correctly at the top-level.

I think if we had iOS tests, I'd recommend renaming iap.test.ts to iap.android.test.ts and creating a separate iap.ios.test.ts that tests iOS logic.

PR #2039, opened 2022-10-11
Fix getSubscriptions Android types; add platform enum and tests

hsource:
## Motivation

When I updated `react-native-iap` to 12.0.3 in my project, I realized that `getSubscriptions` on Android no longer returns anything resembling what the typings say they do.

Instead, I got back output that was very different on iOS and Android.

```ts
    const subscription = {
      subscriptionOfferDetails: [
        {
          pricingPhases: {
            pricingPhaseList: [
              {
                recurrenceMode: 1,
                priceAmountMicros: '49990000',
                billingCycleCount: 0,
                billingPeriod: 'P1Y',
                priceCurrencyCode: 'USD',
                formattedPrice: '$49.99',
              },
            ],
          },
          offerTags: [],
          offerToken: 'dGVzdA==',
        },
      ],
      name: 'MyApp Pro: Annual Plan',
      productType: 'subs',
      description: '',
      title: 'MyApp Pro: Annual Plan (MyApp - Productivity App)',
      productId: 'app_4999_1y_1w0',
    };
```

## Fix

1. Update the Typescript `Subscription` typings to use Union types
2. Since the subscription types are now very different between the different platforms, I updated the logic to add a new additional `platform: SubscriptionPlatform` field to the returned subscriptions. This allows calling code to switch based on them more easily
3. For the RequestPurchase and RequestSubscription types, I also used Union types, and then used ['in' operator narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing) to make the app more type-safe

## Testing

1. Added an automated test to test the typings
2. Created an installable branch with the lib and installed it in my own project using `yarn add wanderlog/react-native-iap#12.0.3-wanderlog.1`
4. Tested fetching subscriptions on a real Android phone

### DIFF
--- a/src/iap.ts
+++ b/src/iap.ts
@@ -757,7 +757,8 @@ export const requestSubscription = (
           if (!('sku' in request)) {
             throw new Error('sku is required for Amazon subscriptions');
           }
-          return RNIapAmazonModule.buyItemByType(request.sku);
+          const {sku} = request;
+          return RNIapAmazonModule.buyItemByType(sku);
         } else {
           if (
             !('subscriptionOffers' in request) ||
--- a/src/iap.ts
+++ b/src/iap.ts
@@ -555,23 +555,24 @@ const App = () => {
 
  */
 
-export const requestPurchase = ({
-  sku,
-  andDangerouslyFinishTransactionAutomaticallyIOS = false,
-  obfuscatedAccountIdAndroid,
-  obfuscatedProfileIdAndroid,
-  appAccountToken,
-  skus, // Android Billing V5
-  isOfferPersonalized = undefined, // Android Billing V5
-  quantity,
-  withOffer,
-}: RequestPurchase): Promise<ProductPurchase | void> =>
+export const requestPurchase = (
+  request: RequestPurchase,
+): Promise<ProductPurchase | void> =>
   (
     Platform.select({
       ios: async () => {
-        if (!sku) {
-          return Promise.reject(new Error('sku is required for iOS purchase'));
+        if (!('sku' in request)) {
+          throw new Error('sku is required for iOS purchase');
         }
+
+        const {
+          sku,
+          andDangerouslyFinishTransactionAutomaticallyIOS = false,
+          appAccountToken,
+          quantity,
+          withOffer,
+        } = request;
+
         if (andDangerouslyFinishTransactionAutomaticallyIOS) {
           console.warn(
             'You are dangerously allowing react-native-iap to finish your transaction automatically. You should set andDangerouslyFinishTransactionAutomatically to false when calling requestPurchase and call finishTransaction manually when you have delivered the purchased goods to the user. It defaults to true to provide backwards compatibility. Will default to false in version 4.0.0.',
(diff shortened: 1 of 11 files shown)


src/types/index.ts
```diff
@@ -172,7 +197,11 @@ export interface SubscriptionIOS extends ProductCommon {
   subscriptionPeriodUnitIOS?: SubscriptionIosPeriod;
 }
 
-export type Subscription = SubscriptionAndroid & SubscriptionIOS;
+export type Subscription =
```
andresesfm, line 200: @hsource The downside I saw to using `|` instead of `&` is that then you'll have to check the type. I had the idea of eventually converting the type to something more like:
```
export type Subscription = {
android: SubscriptionAndroid,
amazon:SubscriptionAmazon,
ios: SubscriptionIOS,
}
```
I'm very curious to read your thoughts

hsource: @andresesfm 

I think the main benefit of this is that `Subscription` will only ever be one of `SubscriptionAndroid` or `SubscriptionIOS`. This lets users more easily have `if` and `switch` statements that work for both iOS and Android.

The documentation on this is here: [discriminated unions](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions)

Here's an example in my real code code:

```ts
function getLocalizedPriceFromSubscription(subscription: Subscription) {
  switch (subscription.platform) {
    case SubscriptionPlatform.android: {
      // Typescript uses discriminated unions to know that the subscription is a SubscriptionAndroid
      const pricingPhaseInfo = getPaidPricingPhaseInfo(subscription);
      return pricingPhaseInfo.formattedPrice;
    }
    case SubscriptionPlatform.ios:
      // Typescript uses discriminated unions to know that the subscription is a SubscriptionIos
      return subscription.localizedPrice;
    case SubscriptionPlatform.amazon:
      throw new Error(
        'getLocalizedPriceFromSubscription: Amazon is currently not supported',
      );
  }
}
```

This makes it much easier to work with the subscription types. This way, we can mark less properties as optional too, which makes working with the types easier.

hsource:
I realize this is causing some typechecking to fail with the example app so will work on fixing that.

hsource:
@andresesfm I updated this to fix the typings. I also adjusted the typings for `requestSubscription` and `requestPurchase` to use intersection types. I added some comments about how this works.

src/iap.ts
```diff
@@ -513,23 +555,24 @@ const App = () => {
...
-}: RequestPurchase): Promise<ProductPurchase | void> =>
+export const requestPurchase = (
+  request: RequestPurchase,
+): Promise<ProductPurchase | void> =>
   (
     Platform.select({
       ios: async () => {
-        if (!sku) {
-          return Promise.reject(new Error('sku is required for iOS purchase'));
+        if (!('sku' in request)) {
```
hsource, line 564: We use the type narrowing technique ([docs](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing)) to tell if this is a `RequestPurchaseIOS` or `RequestPurchaseAndroid`. Typescript knows that only `RequestPurchaseIOS` has the property `sku`, so knows that it's the iOS type after this check.

src/iap.ts
```diff
@@ -513,23 +555,24 @@ const App = () => {
...
+          throw new Error('sku is required for iOS purchase');
```
hsource, line 565: I replaced `return Promise.reject` with `throw`, since those are equivalent in an async function but throw is a bit safer, since we can occasionally forget to include `return`.

src/types/index.ts
hsource: If we converted the type to this:

```ts
export type Subscription = {
  android?: SubscriptionAndroid,
  amazon?: SubscriptionAmazon,
  ios?: SubscriptionIOS,
}
```

The user would still also have to check the type by doing `if (subscription.android) { ... } else if (subscription.ios) { ... }`

I think using union types is a bit clearer because it makes it clear that only one of the 3 will be filled at a time.

tsconfig.json
```diff
@@ -25,5 +25,5 @@
     "strict": true,
     "target": "esnext"
   },
-  "exclude": ["IapExample", "lib"]
+  "exclude": ["lib"]
```
hsource, line 28: I realized this isn't needed, so reverted this.

src/types/index.ts
hsource: If we converted the type to this:

```ts
export type Subscription = {
  android?: SubscriptionAndroid,
  amazon?: SubscriptionAmazon,
  ios?: SubscriptionIOS,
}
```

The user would still also have to check the type by doing `if (subscription.android) { ... } else if (subscription.ios) { ... }`

I think using union types is a bit clearer because it makes it clear that only one of the 3 will be filled at a time.

src/types/index.ts
```diff
@@ -194,7 +223,13 @@ export interface RequestPurchaseIOS {
   withOffer?: Apple.PaymentDiscount;
 }
 
-export type RequestPurchase = RequestPurchaseAndroid & RequestPurchaseIOS;
+/** As of 2022-10-12, we only use the `sku` field for Amazon purchases */
```
andresesfm, line 226: We also use it for iOS. We only use `skus` on Google Play

andresesfm:
Thank you for your contributions @hsource

PR #2058, opened 2022-10-19
Added support for showManageSubscriptions on Storekit 2

timbrandin:
This PR adds support for showManageSubscriptions: https://developer.apple.com/documentation/storekit/appstore/3803198-showmanagesubscriptions

- This panel is very useful for testing where one can cancel the subscription and avoid waiting for expire and additional renewals.
- This panel is very user friendly allowing users to cancel or change their subscription without going to the App Store or Settings app to find their subscription.

### DIFF
--- a/ios/RNIapIosSk2.swift
+++ b/ios/RNIapIosSk2.swift
@@ -85,6 +85,11 @@ protocol Sk2Delegate {
         reject: @escaping RCTPromiseRejectBlock
     )
 
+    func  showManageSubscriptions(
+        _ resolve: @escaping RCTPromiseResolveBlock,
+        reject: @escaping RCTPromiseRejectBlock
+    )
+
     func clearTransaction(
         _ resolve: @escaping RCTPromiseResolveBlock,
         reject: @escaping RCTPromiseRejectBlock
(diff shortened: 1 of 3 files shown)


ios/RNIapIosSk2.swift
```diff
@@ -848,6 +867,30 @@ class RNIapIosSk2iOS15: Sk2Delegate {
         #endif
     }
 
+    @objc public func showManageSubscriptions(
+        _ resolve: @escaping RCTPromiseResolveBlock = { _ in },
+        reject: @escaping RCTPromiseRejectBlock = { _, _, _ in }
+    ) {
+        #if !os(tvOS)
+        guard let scene = UIApplication.shared.connectedScenes.first as? UIWindowScene,
```
andresesfm, line 875: Is this a better way of getting the scene? I was using `await UIApplication.shared.keyWindow?.windowScene` on `beginRefundRequest` line :915

timbrandin: Not sure, maybe this is faster as it isn't async? But I'm not a Swift/Objective-C expert

kyle-ssg:
I was looking around for this functionality. If I'm not mistaken this PR does not actually expose this function in iosSk2.js ?

With that being the case, it looks like people can just achieve this at the moment by doing 

```
   const { RNIapIosSk2 } = NativeModules

   RNIapIosSk2.showManageSubscriptions().catch(() => {
      Linking.openURL('https://apps.apple.com/account/subscriptions')
    })
```

andresesfm:
Addressed your request on https://github.com/dooboolab/react-native-iap/pull/2376 @kyle-ssg

andresesfm:
released as `12.10.2`

kyle-ssg:
Amazing

PR #2112, opened 2022-11-14
Fix compatibility with babel-plugin-module-resolver

jamesedmonston:
Hey,

I'm using `babel-plugin-module-resolver` for import aliases, but doing so is breaking compatibility with this package.

It's failing when trying to import from `'.'` (Metro is trying to import from `../../../../(rootDir)`) – changing this to `'./'` has fixed the issue for me.

Thanks!

### DIFF
--- a/src/types/appleSk2.ts
+++ b/src/types/appleSk2.ts
@@ -5,9 +5,9 @@ import type {
   Purchase,
   SubscriptionIOS,
   SubscriptionIosPeriod,
-} from '.';
+} from './';
 import type * as Apple from './apple';
-import {SubscriptionPlatform} from '.';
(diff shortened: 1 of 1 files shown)


andresesfm:
Hello @jamesedmonston thank you for the PR. Can you please provide any link to documentation that supports this change? I see that https://www.npmjs.com/package/babel-plugin-module-resolver was last updated 2 years ago, so I'm not sure if there's a more recent alternative to it.

hyochan:
I think it may also depend on the OS. @jamesedmonston  Could you kindly provide additional info related to your environment?

hyochan:
Closes this due to inactivity.

PR #2220, opened 2023-01-30
fix: include expo plugin in package.json files

pafry7:
Currently, the `plugin/build` and `app.plugin.js` directories are missing from the package installed via npm. I forgot to add them to the `files` array in package.json 🤦‍♂️


andresesfm:
Thank you @pafry7

PR #2222, opened 2023-01-31
fix: types path

pafry7:
Path to `index.d.ts` changed after adding the plugin.


pavelbabenko:
Fix of:
https://github.com/dooboolab/react-native-iap/issues/2224

andresesfm:
@pafry7 @pavelbabenko can you please explain how is it that adding the plugin changed the path? Would it be better to revert it to point to the old path?

PR #2244, opened 2023-02-10
[Android] subscriptionOfferDetails - added "basePlanId" & "offerId" fields

Mkolakoglu:
At the moment, "basePlanId" and "offerId" fields are missing in subscriptionOfferDetails. Those are very important fields when filtering the offers in complex subscription structures. 

See android docs: https://developer.android.com/reference/com/android/billingclient/api/ProductDetails.SubscriptionOfferDetails

PS:[ react-native-iap docs](https://react-native-iap.dooboolab.com/docs/api/interfaces/SubscriptionOfferAndroid) need to be updated after this change.


andresesfm:
unit tests are failing because the new properties are not added

Mkolakoglu:
> unit tests are failing because the new properties are not added

Updated 👍

andresesfm (approved):

PR #2258, opened 2023-02-22
fix: case of user cancelled case

kaancembertas:
This fix allows developers to know whether user has cancelled the purchase or not when using StoreKit2.


andresesfm:
Looks good, CI doesn't pass because of styling (spaces)

andresesfm (approved):

PR #2264, opened 2023-02-25
Fix place of addTransactionObserver

kaancembertas:
Starting transaction listener at constructor causes unfinished transactions to be missed. This improvement allows developers to start listening transactions right time (when JS Thread bootstraps and then IAP module initialized)

### DIFF
--- a/ios/RNIapIosSk2.swift
+++ b/ios/RNIapIosSk2.swift
@@ -433,7 +433,6 @@ class RNIapIosSk2iOS15: Sk2Delegate {
         self.sendEvent = sendEvent
         products = [String: Product]()
         transactions = [String: Transaction]()
-        addTransactionObserver()
     }
 
     deinit {


andresesfm:
@kaancembertas so where should `addTransactionObserver` be called then?

kaancembertas:
Currently addTransactionObserver function is called at initConnection function.

andresesfm:
If I remember correctly, it was added there to handle the following case: https://developer.apple.com/documentation/storekit/in-app_purchase/original_api_for_in-app_purchase/promoting_in-app_purchases. I'm thinking that by the time the js engine is bootstrapped, it's too late to catch that case, any thoughts on that? 

Note: It is particularly hard to tell if this is the way we are supposed to handle promoted items as there's no updated documentation for SK2

andresesfm:
@kaancembertas Could you please create a ticket with the details of the issue and how this change fixes it? What you mention makes sense but there's little context and no clear way to test if this fixes the issue.

Thank you for your contribution to the project

zarv1k:
Hey guys!

I'm experiencing the same issue - in Sk2 the lib is started listening transactions to early. Pending transactions are emmited before the JS engine bootstrapped, so transaction listener (that is set up in withIAPContext) is quite useless for getting transactions at the start of the app.

This PR at the moment (just removing `addTransactionObserver()` call from the constructor of the `RNIapIosSk2iOS15`) does not fix the issue.

> Currently addTransactionObserver function is called at initConnection function @andresesfm

I think that starting to listen to pending (`Transaction.updates`) transactions at initConnection on iOS native side is the correct place, but it's not enough to fix the issue entirely, because the transaction listener in `withIAPContext` for now is setting up only after the initConnection() promise resolve and even after the `setState(connected)` in useEffect hook.

I think the transaction listener on the JS side in `withIAPContext` should be setup before calling `initConnection()` to prevent some race conditions where we can miss some piece of pending transactions on the early init app stages.

kaancembertas:
Hello again!

Is there a solution if we do this? We can still call addTransactionObserver from constructor and we can create a cache variable to store transaction updates which could not be received by js-side. After a listener added from js side, we can send the cache to js side.

andresesfm:
Based on the history of the sk1 implementation. I have the feeling that by the time the JS engine boots up, it's already too late to capture the promotion purchase.

zarv1k:
Hey @andresesfm , @kaancembertas!

FYI, Using SK2 with `react-native-iap@12.8.3` I still have missing transactions on JS side - I see them in swift `Transaction.updates` queue from but they are not sent to JS side, because at that moment `RNIapIosSk2iOS15.hasLIsteners` is `false`.

I'm definitely sure that is because the `addTransactionObserver()` calling in `initConnection()` - at this point of time there is no listeners ready for handling the emitted transactions.

Class `RNIapIosSk2` implements `RCTEventEmitter` interface. If we take a look at `RCTEventEmitter` there is a comment:

Screenshot 2023-03-17 at 12 06 14

I'm pretty sure that the call of `addTransactionObserver()` should be done at `startObserving` - it's guaranteed that the listener on JS side is ready to handle transaction events from the native side. 

And it might be also worth to call the `removeTransactionObserver()` in `startObserving()`.

I could create PR for that. 

What do you think?

zarv1k:
#2344

andresesfm:
Thank you @zarv1k  merged your PR and released a new version

PR #2281, opened 2023-03-13
feat: StoreKit2 - add environment property into TransactionSk2

zarv1k:


src/types/appleSk2.ts
```diff
@@ -96,6 +96,7 @@ export type TransactionSk2 = {
   deviceVerification: string;
   deviceVerificationNonce: string;
   expirationDate: number;
+  environment: 'Production' | 'Sandbox' | 'Xcode' | null;
```
andresesfm, line 99: Why null and not undefined/ optional. I think we are optional fields everywhere else

andresesfm: Can you please add a documentation comment saying that it's only available on iOS 16?

Thank you for your contribution

zarv1k: I don't think it's only available in iOS 16, coz the swift implementation in serialize helper in my PR is using either `transaction.environment` (which is stable since iOS 16) or `transaction.environmentStringRepresentation` (which is available since iOS 15 but in some rare cases it could return not an expected environment string according to docs in StoreKit from XCode - see attachments for details)

ios/IapSerializationUtils.swift
```diff
@@ -138,12 +138,24 @@ func serialize(_ ot: Product.SubscriptionOffer.OfferType?) -> String? {
 }
 @available(iOS 15.0, tvOS 15.0, *)
 func serialize(_ t: Transaction) -> [String: Any?] {
+    var environment: String?
+
+    if #available(iOS 16.0, *) {
```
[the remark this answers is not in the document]
zarv1k: Good point, just fixed that

PR #2413, opened 2023-04-17
feat: include JWS representation of purchase verification when using StoreKit 2

oscnord:
This PR adds the JWS representation of the App Store verification when using StoreKit 2 (iOS/tvOS) to the purchase response as well as `purchase-updated` and `iap-transaction-updated` events.

### DIFF
--- a/ios/IapSerializationUtils.swift
+++ b/ios/IapSerializationUtils.swift
@@ -176,6 +176,12 @@ func serialize(_ t: Transaction) -> [String: Any?] {
     ]
 }
 @available(iOS 15.0, tvOS 15.0, *)
+func serialize(_ t: Transaction, _ v: VerificationResult<Transaction>) -> [String: Any?] {
+    var transaction = serialize(t)
+    transaction.updateValue(v.jwsRepresentation, forKey: "verificationResult")
+    return transaction
+}
+@available(iOS 15.0, tvOS 15.0, *)
 func serialize(_ ot: Transaction.OfferType?) -> String? {
     guard let ot = ot else {return nil}
     switch ot {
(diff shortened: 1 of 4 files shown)


ios/IapSerializationUtils.swift
```diff
@@ -176,6 +176,10 @@ func serialize(_ t: Transaction) -> [String: Any?] {
     ]
 }
 @available(iOS 15.0, tvOS 15.0, *)
+func serialize(_ t: Transaction, _ v: VerificationResult<Transaction>) -> [String: Any?] {
+    return ["verificationResult": v.jwsRepresentation].merging(serialize(t)) { (current, _) in current }
+}
```
oscnord, line 181: This could probably be handled in a better way by keeping the verification and transaction separate and skip a merge. However I wanted to avoid introducing new events/bigger changes to the code base. 

If there is a better approach to include the verification result then let me know.

andresesfm: Wouldn't it simply be `serialize(t).updateValue(value: v.jwsRepresentation, forKey:"verificationResult")`?

ref: https://stackoverflow.com/a/32249397/570612

oscnord: Looks like it needs to be something like:
```swift
@available(iOS 15.0, tvOS 15.0, *)
func serialize(_ t: Transaction, _ v: VerificationResult<Transaction>) -> [String: Any?] {
    var transaction = serialize(t)
    return transaction.updateValue(v.jwsRepresentation, forKey:"verificationResult") as! [String : Any?]
}
```
Due to to the need of typecasting and resulting immutable value.

andresesfm: I think this should work and it's more readable: 
```
@available(iOS 15.0, tvOS 15.0, *)
func serialize(_ t: Transaction, _ v: VerificationResult<Transaction>) -> [String: Any?] {
    var transaction = serialize(t)
    transaction.updateValue(v.jwsRepresentation, forKey:"verificationResult")
    return transaction
}
```

Does that work for you?

oscnord: Yes that works!

andresesfm:
Merged and released. Thank you @oscnord

PR #2459, opened 2023-06-14
docs: purchases.md use object instead of params in requestSubscription call

frothAml:
The example shown in purchases.md has the wrong interface use for the function `requestSubscription`

In the docs example is shown as
```typescript
      await requestSubscription(
        {sku},
        ...(offerToken && {subscriptionOffers: [{sku, offerToken}]}),
      );
```

While the interface to the function requires an object of the type `RequestSubscription` and not 2 args


hyochan (approved):
LGTM!

Issue #1696, opened 2022-05-15
iOS: Many crashes due to lack of thread-safety


mars-lan:
> Please use [Discussion board](https://github.com/dooboolab/react-native-iap/discussions) if you want get some help out of it. Please report issue if there is a bug.

### Version of react-native-iap

8.0.8

### Version of react-native

0.67.4

### Platforms you faced the error (IOS or Android or both?)

iOS

### Expected behavior

No crashing

### Actual behavior

Random crashes with stack trace like this:

```
Crashed: com.facebook.react.RNIapIosQueue
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000000

<compiler-generated> - Line 388412 RNIapIos.addPromise(forKey:resolve:reject:) + 388412
RNIapIos.swift - Line 165 RNIapIos.getItems(_:resolve:reject:) + 165
<compiler-generated> - Line 391604 @objc RNIapIos.getItems(_:resolve:reject:) + 391604
```

### Tested environment (Emulator? Real Device?)

Real device

### Steps to reproduce the behavior

This is caused by non-thread-safe access to the dictionary `promisesByKey`. A [serial queue](https://github.com/dooboolab/react-native-iap/blob/50ab0d0d936307daf55a9ab9b2dce05cb1c6825f/ios/RNIapIos.swift#L47) was added to protect some of these accesses but not all. IMHO a better way to protect this is by adding `dispatch` directly inside of `addPromise`, `resolvePromises` & `rejectPromises`.

edodusi:
I have those crashes too with `8.1.3` and also this error which I think it's related:

```
EXC_BAD_ACCESS KERN_PROTECTION_FAILURE 0x0000005f774d5340

...
0x5ed518 RNIapIos.resolvePromises(forKey:value:) + 100 (RNIapIos.swift:100)
0x5f59f0 RNIapIos.productsRequest(_:didReceive:) + 446 (RNIapIos.swift:446)
0x5f5bb8 @objc RNIapIos.productsRequest(_:didReceive:) + 68 (<compiler-generated>:68)
0x3a18 __27-[SKProductsRequest _start]_block_invoke_2 + 188
...
```

edodusi:
(also this should be labeled as bug I think)

aadityapaliwal94:
Can anyone look into this issue? Causing crash on production.

andresesfm:
[...]

PR #2518, opened 2023-08-15
fix: race condition crash

arthurgeron:
Doing `getProduct` on iOS would sometimes cause a crash due to a racing condition. This fixes it.
Basically this change makes the products' store and listeners thread-safe.

### DIFF
--- a/ios/RNIapIosSk2.swift
+++ b/ios/RNIapIosSk2.swift
@@ -424,20 +424,57 @@ class RNIapIosSk2: RCTEventEmitter, Sk2Delegate {
 
 @available(iOS 15.0, tvOS 15.0, *)
 class RNIapIosSk2iOS15: Sk2Delegate {
-    private var hasListeners = false
-    private var products: [String: Product]
+    private var _hasListeners = false
+    private let _hasListenersQueue = DispatchQueue(label: "com.dooboolab.rniap.hasListenersQueue", attributes: .concurrent)
     private var transactions: [String: Transaction]
     private var updateListenerTask: Task<Void, Error>?
     fileprivate var sendEvent: ((String?, Any?) -> Void)?
+    var hasListeners: Bool {
+        get {
+            return _hasListenersQueue.sync { _hasListeners }
+        }
+        set {
+            _hasListenersQueue.async(flags: .barrier) { [weak self] in 
+                self?._hasListeners = newValue 
+            }
+        }
+    }
     init(sendEvent: ((String?, Any?) -> Void)? ) {
         self.sendEvent = sendEvent
-        products = [String: Product]()
         transactions = [String: Transaction]()
     }
 
     deinit {
         removeTransactionObserver()
     }
+
+    actor ProductStore {
+        private(set) var products: [String: Product] = [:]
+
+        func addProduct(_ product: Product) {
+            self.products[product.id] = product
+        }
+
+        func getAllProducts() -> [Product] {
+            return Array(self.products.values)
+        }
+
+        func getProduct(productID: String) -> Product? {
+            return self.products[productID]
+        }
+
+        func removeAll() {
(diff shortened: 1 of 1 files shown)


matthargett:
can you reproduce the crash in a unit/integration test, that fails without the fix and passes with the fix?

arthurgeron:
> can you reproduce the crash in a unit/integration test, that fails without the fix and passes with the fix?

I'll try to, but I can't give an ETA,  although since getProduct is async, trying to access/write in rapid sequence will inevitably lead to a crash due to a racing condition, which does not occur on Android.

I'll also revert my local changes so I can post the crash logs here.

arthurgeron-work:
> can you reproduce the crash in a unit/integration test, that fails without the fix and passes with the fix?

With the current testing structure, it's impossible to reproduce this crash within unit/integration tests.

hyochan (approved):
I haven't been able to thoroughly test the changes made in this PR, but trusting ChatGPT's review seems like a good idea

Screenshot 2023-11-01 at 5 45 50 AM

arthurgeron-work:
@hyochan That's a great way of approaching code review on this scenario. Do you mind sharing which format you've used to show code changes to ChatGPT?

geroale:
Hello, this pull didn't fix the problem for us. We are still getting https://github.com/dooboolab-community/react-native-iap/issues/1696

Yann-prak:
@geroale The fix was only on the `RNIapIosSk2.swift` file. If you are still using the STOREKIT1 option like I do, the error occurs on the file `RNIapIos.swift`. You have to apply the same fix applied in the commit. I'm not a swift guy so I don't really know how to fix the thread safety for Storekit 1 😕

arthurgeron-work:
> @geroale The fix was only on the `RNIapIosSk2.swift` file. If you are still using the STOREKIT1 option like I do, the error occurs on the file `RNIapIos.swift`. You have to apply the same fix applied in the commit. I'm not a swift guy so I don't really know how to fix the thread safety for Storekit 1 😕

@geroale @Yann-prak  I'd like to help but I need more stack traces, If it's the same error as #1696 you could post there, or create a new issue if it's different. 

A workaround I had found for my crashes was to never run more than one getProducts or getSubscriptions call at the same time, essentially adding a loading state.

arthurgeron:
@geroale @Yann-prak  I've done a somewhat similar fix to Store Kit 1 code.

geroale:
@arthurgeron Thank you. We had also other concurrency problems in Store Kit 1 by accessing `promisesByKey` in the method `resolvePromises`. Because of this, we wrapped `resolvePromises` with `myQueue.sync` https://github.com/Spicy-Sparks/react-native-iap/blob/main/ios/RNIapIos.swift

From what I'm seeing, your and our changes are compatible, but is there a better way to integrate them, maybe reworking our ones?

arthurgeron:
@geroale Possibly better to bring over your changes, I'll take a look and try later!

Yann-prak:
@arthurgeron you can find a full discussion there #2550 with some stack traces.

Yesterday I've done a similar fix with a ProductStore in local and it was working. I did the same patch as @geroale too. I had some memory issues but I'm not sure if it was linked to the patches or to something else as I did not develop in swift for a long time now. I can share my code if needed.

marf:
> @geroale Possibly better to bring over your changes, I'll take a look and try later!

@arthurgeron we tested our changes at https://github.com/Spicy-Sparks/react-native-iap/blob/main/ios/RNIapIos.swift and the concurrency crashes seem to have disappeared.

eight-buddha:
I confirm that the @geroale  fix works. I don’t understand why one application with the same version of the library has no problems, but the other constantly crashes. As I understand it, this fix will still not be added to the library. Currently using 12.13.2 version.

PR #2532, opened 2023-08-29
fix : edit requestPruchase requestSubscription TS and android return

ddowoo:
While using this library in my project, I checked that requestPurchase and requestSubscription have different return values ​​from TypeScript. And I modified it so that the return value came out according to typescript.

### DIFF
--- a/src/iap.ts
+++ b/src/iap.ts
@@ -576,7 +576,7 @@ const App = () => {
 
 export const requestPurchase = (
   request: RequestPurchase,
-): Promise<ProductPurchase | ProductPurchase[] | void> =>
+): Promise<ProductPurchase | void> =>
   (
     Platform.select({
       ios: async () => {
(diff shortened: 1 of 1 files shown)


src/iap.ts
```diff
@@ -576,7 +576,7 @@ const App = () => {
 
 export const requestPurchase = (
   request: RequestPurchase,
-): Promise<ProductPurchase | ProductPurchase[] | void> =>
+): Promise<ProductPurchase | void> =>
```
hyochan, line 579: @ddowoo The return value could be an array so I think you should revert this changes.

Reference https://developer.apple.com/forums/thread/6431#14831

src/iap.ts
```diff
@@ -732,7 +734,7 @@ const App = () => {
  */
 export const requestSubscription = (
   request: RequestSubscription,
-): Promise<SubscriptionPurchase | SubscriptionPurchase[] | null | void> =>
+): Promise<SubscriptionPurchase |  null | void> =>
```
hyochan, line 737: Same here.

Issue #2535, opened 2023-08-30
Module was compiled with an incompatible version of Kotlin.


17Amir17:
> Please use the [Discussion board](https://github.com/dooboolab-community/react-native-iap/discussions) if you want to get some help. Please use issues to report bugs.

**Description**
Documentation states you should add `kotlinVersion = "1.6.0"`
After updating to `12.10.7` I get `Module was compiled with an incompatible version of Kotlin.`

**Solution**
If anyone else is seeing this either update kotlinVersion to  `1.8.0` or downgrade lib version to `12.10.5`
I'll raise pr to update docs

galenweber:
Updating the kotlinVersion to `1.8.0` I get the following error:
```
Duplicate class kotlin.collections.jdk8.CollectionsJDK8Kt found in modules jetified-kotlin-stdlib-1.8.0 (org.jetbrains.kotlin:kotlin-stdlib:1.8.0) and jetified-kotlin-stdlib-jdk8-1.7.22 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.22)
Duplicate class kotlin.internal.jdk8.JDK8PlatformImplementations found in modules jetified-kotlin-stdlib-1.8.0 (org.jetbrains.kotlin:kotlin-stdlib:1.8.0) and jetified-kotlin-stdlib-jdk8-1.7.22 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.22)
Duplicate class kotlin.internal.jdk8.JDK8PlatformImplementations$ReflectSdkVersion found in modules jetified-kotlin-stdlib-1.8.0 (org.jetbrains.kotlin:kotlin-stdlib:1.8.0) and jetified-kotlin-stdlib-jdk8-1.7.22 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.22)
Duplicate class kotlin.jvm.jdk8.JvmRepeatableKt found in modules jetified-kotlin-stdlib-1.8.0 (org.jetbrains.kotlin:kotlin-stdlib:1.8.0) and jetified-kotlin-stdlib-jdk8-1.7.22 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.22)
[...]

PR #2536, opened 2023-08-30
Update Kotlin Version In Docs  

17Amir17:
#2535
After version `12.10.7` build will fail with
```
INF/java.com.google.android.libraries.play.billing.public.ktbilling_granule.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.8.0, expected version is 1.6.0.
```
Downgrading to `12.10.5` works with 1.6.0


hyochan (approved):
LGTM!

PR #2543, opened 2023-09-14
Upgrade Amazon's SDK to 3.0.4

KonstantinosKoulaxizis:
It's just one patch upgrade, as you can see in the [release notes](https://developer.amazon.com/docs/appstore-sdk/release-notes.html) nothing breaking was introduced so the upgrade will be smooth.

### DIFF
--- a/.github/workflows/ci-example-android.yml
+++ b/.github/workflows/ci-example-android.yml
@@ -25,6 +25,11 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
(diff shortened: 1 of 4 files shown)


hyochan (approved):

hyochan (approved):

hyochan (approved):
I had to add `PurchaseResponse.RequestStatus.PENDING` in order to fix build failing. The new status has been introduced in `3.0.4`

PR #2599, opened 2023-11-03
fix: correct gradle import instructions in get-started.mdx

SufianBabri:
While integrating this on Android, I found that the suggested gradle import changes weren't correct on the [Get Started page](https://react-native-iap.dooboolab.com/docs/get-started#with-androidx).

I have made the correction in this PR.


hyochan (approved):
LGTM!

PR #2614, opened 2023-11-17
feat: add renewal info for Storekit 2

cervebar:
- add renewal info (uncomment and serialize)
- change debugSerialize to serialize - there is info in this property useful for production not only for development
- added transactionReasonIOS enum (to distinguish between type of renewing transactions of purchase events)

see also: [JWSTransactionDecodedPayload](https://developer.apple.com/documentation/appstoreserverapi/jwstransactiondecodedpayload)

### DIFF
--- a/ios/RNIapIosSk2.swift
+++ b/ios/RNIapIosSk2.swift
@@ -277,10 +277,6 @@ class RNIapIosSk2: RCTEventEmitter, Sk2Delegate {
         delegate.stopObserving()
     }
 
-    override func addListener(_ eventName: String?) {
-        super.addListener(eventName)
-    }
-    
     /**
      "iap-transaction-updated" is unique to Sk2.
      "iap-promoted-product" is only avaiable on Sk1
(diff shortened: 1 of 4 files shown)


ios/RNIapIosSk2.swift
```diff
@@ -277,10 +277,6 @@ class RNIapIosSk2: RCTEventEmitter, Sk2Delegate {
         delegate.stopObserving()
     }
 
-    override func addListener(_ eventName: String?) {
```
cervebar, line 280: code review note: 
- not my intention
- this was removed by `yarn swift:lint`, otherwise the lint is failing

src/types/appleSk2.ts
```diff
@@ -134,8 +134,22 @@ export type SubscriptionStatus =
   | 'revoked'
   | 'subscribed';
 
+/**
+* Renewal info for whole subscription group.
+* see: https://developer.apple.com/documentation/storekit/product/subscriptioninfo/status/3822294-renewalinfo
+* WARN:
+* - autoRenewPreference is serialised as autoRenewProductId in jsonRepresentation
+* - renewalDate is available in jsonRepresentation (will change with Xcode 15 https://developer.apple.com/forums/thread/738833)
```
cervebar, line 142: once we will update to Xcode 15, then we will be able to get renewal date form new API , see Apple answer here: https://developer.apple.com/forums/thread/738833

hyochan (approved):
LGTM!

PR #2616, opened 2023-11-18
feat: add ios error reason user canceled

cervebar:
- adding user canceled error code, so we can distinguish in app what error is intention and what is real error

### DIFF
--- a/ios/RNIapIosSk2.swift
+++ b/ios/RNIapIosSk2.swift
@@ -277,10 +277,6 @@ class RNIapIosSk2: RCTEventEmitter, Sk2Delegate {
         delegate.stopObserving()
     }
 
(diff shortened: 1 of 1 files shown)


ios/RNIapIosSk2.swift
```diff
@@ -277,10 +277,6 @@ class RNIapIosSk2: RCTEventEmitter, Sk2Delegate {
         delegate.stopObserving()
     }
 
-    override func addListener(_ eventName: String?) {
```
cervebar, line 280: code review note:

not my intention
this was removed by yarn swift:lint, otherwise the lint is failing

hyochan (approved):
LGTM!

PR #2630, opened 2023-11-27
fix: add missing conn for SK2 intro offers

Shaw-Signaturize:
Adding missing mapping for intro offers in storekit2 on existing type

### DIFF
--- a/src/types/appleSk2.ts
+++ b/src/types/appleSk2.ts
@@ -52,7 +52,7 @@ export const productSk2Map = ({
   description,
   displayName,
   price,
-   currency,
+  currency,
   displayPrice,
 }: ProductSk2): ProductIOS => {
   const prod: ProductIOS = {
@@ -62,7 +62,7 @@ export const productSk2Map = ({
     type: 'iap',
     price: String(price),
     localizedPrice: displayPrice,
-    currency
+    currency,
   };
   return prod;
 };
@@ -88,6 +88,17 @@ export const subscriptionSk2Map = ({
     subscriptionPeriodNumberIOS: `${subscription?.subscriptionPeriod?.value}`,
     subscriptionPeriodUnitIOS:
       subscription?.subscriptionPeriod?.unit.toUpperCase() as SubscriptionIosPeriod,
+    introductoryPriceAsAmountIOS: subscription?.introductoryOffer?.displayPrice,
+    introductoryPricePaymentModeIOS:
+      subscription?.introductoryOffer?.paymentMode.toUpperCase() as
+        | ''
+        | 'FREETRIAL'
+        | 'PAYASYOUGO'
+        | 'PAYUPFRONT',
+    introductoryPriceNumberOfPeriodsIOS:
+      subscription?.introductoryOffer?.period?.value?.toString(),
+    introductoryPriceSubscriptionPeriodIOS: subscription?.introductoryOffer
+      ?.period?.unit as SubscriptionIosPeriod,
   };
   return prod;
 };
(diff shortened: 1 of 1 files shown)


Shaw-Signaturize:
Not sure who to ping for this but it seems the failing check is on a diff of swift files but none were changed in this PR. 
Could this be an issue in the check logic or is there something i have neglected to do.

LectricAvenue:
@Shaw-Signaturize, thanks so much for this, I ran into this myself and came up with a similar solution, but I wasn't sure if it was a result of not using RNIAP as intended.

The type for SubscriptionIOS looks like this:

```javascript
export interface SubscriptionIOS extends ProductCommon {
  platform: SubscriptionPlatform.ios;
  type: 'subs';
  discounts?: Discount[];
  introductoryPrice?: string;
  introductoryPriceAsAmountIOS?: string;
  introductoryPricePaymentModeIOS?:
    | ''
    | 'FREETRIAL'
    | 'PAYASYOUGO'
    | 'PAYUPFRONT';
  introductoryPriceNumberOfPeriodsIOS?: string;
  introductoryPriceSubscriptionPeriodIOS?: SubscriptionIosPeriod;
  subscriptionPeriodNumberIOS?: string;
  subscriptionPeriodUnitIOS?: SubscriptionIosPeriod;
}
```

I think what you have assigned to `introductoryPriceAsAmountIOS` should actually be assigned to `introductoryPrice`.

`introductoryPriceAsAmountIOS` should instead be the value without the currency symbol in front.

I believe that would make the proper change the following:

```javascript
{
  // unchanged fields above
  introductoryPrice: subscription?.introductoryOffer?.displayPrice,
  introductoryPriceAsAmountIOS: subscription?.introductoryOffer?.price.toString(),
  introductoryPricePaymentModeIOS:
    subscription?.introductoryOffer?.paymentMode.toUpperCase() as
      | ''
      | 'FREETRIAL'
      | 'PAYASYOUGO'
      | 'PAYUPFRONT',
  introductoryPriceNumberOfPeriodsIOS:
    subscription?.introductoryOffer?.period?.value?.toString(),
  introductoryPriceSubscriptionPeriodIOS: subscription?.introductoryOffer
    ?.period?.unit as SubscriptionIosPeriod,
}
```

Since the library also returns `currency: ''` with a note that the field isn't available with SK2. I used to get `countryCode` and `discounts` with SK1, so maybe it would be good to add those in with similar notes so that it is easier for people to find in the future.

Shaw-Signaturize:
@LectricAvenue
You could very well be right, i did have a look at the type before adding the fields but as some of the fields don't match exactly so it required a bit of creative license. 
I was originally hoping to get a bit of guidance from the project maintainers to make sure the mappings were what they had envisioned as with them being on the type and on the sk2 response it just seemed like a slight oversight that they were missed from the mapping.

LectricAvenue:
@Shaw-Signaturize of course! I agree that potentially they intended to remove these, although surely they didn't mean to leave out the intro offer details entirely.

The docs seem very outdated, but in the source code I see lots of documentation markup that doesn't seem to be available on the website?

hyochan (changes requested):
Could you kindly check ci failing?

Shaw-Signaturize:
> Could you kindly check ci failing?

@hyochan 

As mentioned above the CI seems to be failing when verifying that certain swift files have not changed. 
The CI is under the impression that file `ios/RNIapIosSk2.swift` has changed but no modifications have been made to it in this PR. 
I was unsure if this was an issue with the CI or there was something i have neglected to do

hyochan (approved):
I've rebased codebase myself~
Let's see how this code works!

Shaw-Signaturize:
@hyochan 

Thanks. The help is much appreciated

PR #2664, opened 2024-01-04
feat: Added export finishTransaction() from StoreKit2.

khrulev:
Added a previously missing exported function that allows manual completion of transactions in **StoreKit2**.

Function has already been described in the interface of the **IosModulePropsSk2** in the codebase but lacked accessible invocation.

The function is crucial and must be used to manually complete transactions in in-app purchases using **StoreKit2**.


hyochan (approved):
Thanks it looks promising!

PR #2669, opened 2024-01-09
fix: avoid app crash due to multiple reject invocation

khrulev:
Probably, someone forgot to delete the old **Reject** when adding the condition to the code. This leads to an error and crash of the application in the development environment. 

**Steps To Reproduce**:
The error occurs when the user has canceled iOS system prompt for Sync with App Store using StoreKit2.

**Error Message:**
The error says: 'Illegal callback invocation from native module. This callback type only permits a single invocation from native code.'

**Solution**:
Delete redundant reject.


hyochan (approved):
Thanks for the fixes!

PR #2681, opened 2024-01-30
Update troubleshooting.mdx

Vittor-Javidan:
When I implemented a subscription system on my app, I tried for a week to find why my payments always got refused inside my app, because i could not find anything wrong with my source code. Until i realize there was a hold on my payment profile, meaning google play by default will refuses any payment attempt, even for test environments.

I figure out this issue could be one that can frustrate a lot of people, and it's not the lib fault.


hyochan (approved):
LGTM!

PR #2687, opened 2024-02-16
Update hooks.md documentation

emilcieslar:
The documentation of getProducts hook was outdated.

If you agree, I'd also add more comprehensive documentation when it comes to handling errors. I was getting an empty products list and without adding the `.catch` statement to the `getProducts` and seeing that I was missing the `skus` property, I couldn't figure it out. I think it should be clear from the documentation that `getProducts` returns a promise.

### DIFF
--- a/docs/docs/api-reference/hooks.md
+++ b/docs/docs/api-reference/hooks.md
@@ -62,7 +62,7 @@ const App = () => {
     <>
       <Button
         title="Get the products"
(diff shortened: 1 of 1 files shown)


hyochan:
@emilcieslar Hi thank you for your contribution! I also agree on your though on adding detailed documents on handling error

emrahyurttutan (approved):

hyochan (approved):

Issue #2701, opened 2024-03-08
It crashes when I launch the app on iPHone6 iOS 12,  but it doesn't crash on iOS 13


zyestin:
**Description**

After the ipa package is generated in archive.
It crashed every time I opened the app, on iPhone 6 / iOS 12. 
However, there's no problem in debug.

> It is possible that the minimum iOS version supported by swift is 13

**Expected Behavior**

App launched normally

**Screenshots**

https://github.com/dooboolab-community/react-native-iap/assets/51897571/ba8b6f5e-5c26-408b-873f-a889d5ae277e

**Environment:**

- react-native-iap: 12.13.0
- react-native: 0.61.5
- Platforms (iOS, Android, emulator, simulator, device): iOS 12

**To Reproduce**

Once the iap library is integrated into any project and archived to generate ipa. 
Every time open the app on an iPhone 6 or iOS 12 machine,  it crashes.

---

[Optional] **Additional Context**

Xcode run with Release, then get it's crash log
```
* thread #1, stop reason = signal SIGABRT
  * frame #0: 0x000000010564e418 dyld`__abort_with_payload + 8
    frame #1: 0x000000010564da58 dyld`abort_with_payload_wrapper_internal + 100
    frame #2: 0x000000010564da88 dyld`abort_with_payload + 12
    frame #3: 0x00000001056108f8 dyld`dyld::halt(char const*) + 304
    frame #4: 0x00000001056136e4 dyld`dyld::_main(macho_header const*, unsigned long, int, char const**, char const**, char const**, unsigned long*) + 5396
    frame #5: 0x000000010560d044 dyld`_dyld_start + 68
```

PR #2702, opened 2024-03-11
Update get-started.mdx to avoid crashing when app launching on iOS 12 

zyestin:
## Summary:

After the ipa package generated by Xcode archive  was installed on on iPhone 6 / iOS 12. It crashed every time I opened the app.

More information can be found here [#2701](https://github.com/dooboolab-community/react-native-iap/issues/2701)

## What was done:

Add a note in get-started.mdx. So developers know to do extra configuration on iOS 12 to avoid crashes


hyochan (approved):
LGTM!

PR #2739, opened 2024-05-10
fix: Send purchase-updated instead of purchase-update.

dwyery:
During some testing of restoring purchases using the getAvailablePurchases method with the alsoPublishToEventListener argument set to true I noticed that the events were not being published correctly. I also noticed in the logs an error message saying that purchase-update was not a supported event.

I had a look at the code in RNIapIosSk2.swift and it appears that the event name should actually be purchase-updated.

I fixed this locally and the error no longer appears and the purchase is now being published to the event listener.

Creating this as a draft PR as I am not too sure if there is anything else I need to do, this is my first contribution to the project.

### DIFF
--- a/.github/workflows/ci-example-ios.yml
+++ b/.github/workflows/ci-example-ios.yml
@@ -48,13 +48,11 @@ jobs:
           bundler-cache: true
           ruby-version: '2.7'
 
-      - name: GitHub Action for SwiftLint (Only files changed in the PR)
-        uses: norio-nomura/action-swiftlint@3.2.1
-        env:
-          DIFF_BASE: ${{ github.base_ref }}
+      - name: Install SwiftLint
+        run: brew install swiftlint
 
-      # - name: SwiftLint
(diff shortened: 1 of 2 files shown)


dwyery:
Looks like the build is failing with:

Run swiftlint lint --fix --format --path ios/*.swift --config .swiftlint.yml

/Users/runner/work/_temp/cc0d47bb-9bc0-431a-bb28-0434ffd4767d.sh: line 1: swiftlint: command not found

Is there something I am missing?

hyochan:
> Looks like the build is failing with:
> 
> Run swiftlint lint --fix --format --path ios/*.swift --config .swiftlint.yml
> 
> /Users/runner/work/_temp/cc0d47bb-9bc0-431a-bb28-0434ffd4767d.sh: line 1: swiftlint: command not found
> 
> Is there something I am missing?

It is weird that swiftlint isn't preinstalled in recent macos image. I've added `brew install swiftlint` step.

hyochan (approved):
LGTM!

PR #2759, opened 2024-06-06
Update troubleshooting.mdx

Vittor-Javidan:
This is an update from commit #2681, now considering any platform.

I decided to update my last contribution after helping a friend identify the root cause of the "getSubscriptions" method returning an empty array on iOS, while it returned products without issue. After about two weeks of exploring numerous possibilities, he finally resolved the problem: some forms on the Apple platform had not been approved yet.

From this experience, I concluded why developers often find it challenging to implement in-app purchases. The main issue is the misleading feedback provided by the platform APIs, which makes it difficult to debug issues logically. Therefore, it is much better for developers to complete any bureaucratic processes before starting to code, allowing them to properly debug other issues.

### DIFF
--- a/docs/docs/guides/troubleshooting.mdx
+++ b/docs/docs/guides/troubleshooting.mdx
@@ -10,13 +10,16 @@ import AdFitTopFixed from "@site/src/uis/AdFitTopFixed";
 
 # Troubleshooting
 
+## Recommendations
+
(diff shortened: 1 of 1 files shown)


hyochan (approved):
Thanks for your warmness 🌤️

PR #2807, opened 2024-07-24
fix: set connected false in endConnection

nej1044:
Hello.
I faced one problem using this.
Though a first purchase after the in-app connection is successful, the store purchase will be completed, but the `currentPurchase` status is not updating inside the library. I would have to repeat the purchase process by killing the app. 
In my opinion, the purchaseListener is connected by the change of the `connected` value inside `useEffect`, which changes from `false` to `true` only when the app is first connected and not after that, so I decided that I can't subscribe to the status of the current order.
So I added a code inside the `endConnection` method to make `connected` false.
This allows you to update every subscription correctly by changing `connected` to `true` when you access the order page and `false` when you access `endConnection`.
Both iOS and Android work as expected.

I think this issue is related to the issue below.
https://github.com/dooboolab-community/react-native-iap/issues/2700

I hope that this PR will be reflected so that it can help everyone.

### DIFF
--- a/src/hooks/useIAP.ts
+++ b/src/hooks/useIAP.ts
@@ -52,6 +52,7 @@ export const useIAP = (): IAP_STATUS => {
     currentPurchase,
     currentPurchaseError,
     initConnectionError,
+    setConnected,
     setProducts,
     setSubscriptions,
     setAvailablePurchases,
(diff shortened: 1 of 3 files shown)


nej1044:
I thought so too.
However, if setConnected (false) is added in the useEffect return, it seems necessary to test it whether its working normally. I'll fix and test it as soon as possible and add a commit once again if both IOS and Android are working fine

nej1044:
Hello, As we discussed before, we have moved setConnected inside useIAP.
However, if you set setConnected to false during just the unmount period, it is not connected again during mount while testing in the test process, which results the connected state becomes false inside context and does not update the fetch status, so we made setConnected true even during mount.

I've tested both iOS and Android, and they both work as expected.

If there are no problems, this update may help with many issues.

hyochan (approved):
This looks very promising! Thanks a lot 👍

PR #2809, opened 2024-07-26
doc(requestPurchase): fix the example in docstrings

Abdelalim-dev:
`useCallback` returns a function (async in this case) that cannot be directly used in the render phase.
Therefore, a state should hold the result of the fetch.


hyochan (approved):
Agree!

PR #2821, opened 2024-08-21
Added showManageSubscriptions to deepLinkToSubsctiptions

benglasgow:
I noticed there wasn't a convenient way to use showManageSubscriptions, so I made this change to make using function that easier.

### DIFF
--- a/src/iap.ts
+++ b/src/iap.ts
@@ -916,7 +916,7 @@ export const deepLinkToSubscriptions = ({
     Platform.select({
       ios: async () => {
         if(isIosStorekit2()) {
-          IapIosSk2.showManageSubscriptions()
+          IapIosSk2.showManageSubscriptions();
         } else{
           IapIos.deepLinkToSubscriptionsIos();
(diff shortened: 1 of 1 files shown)


benglasgow:
This seemed like the most sensible place to add this even though it's not strictly a deeplink.

At first, I started to simply export everything from iosSk2 in modules/index, but then I'd need to make some more changes to what's exported in iosSk2 since it also has a BuyProduct type and finishTransaction which seemed like a bit of a slippery slope. 

Next, I thought about adding it to ios.ts, but then that'd break the pattern of everything related to sk1 being there, and everything related to sk2 being in iosSk2.

Finally, I figured since deepLinkToSubscriptions takes you to essentially the same place as showManageSubscriptions just outside of the app, then it might just make the most sense to add a call to showManageSubscriptions there.

Lmk what you think! It's totally possible that I could be missing something obvious here.

devyeshtandon:
Why did you close this issue?

PR #2849, opened 2024-09-19
Ability to disable amazon's DRM

KonstantinosKoulaxizis:
In some cases, users want to manage the DRM with their own solution and not use Amazon's DRM. In the past, this was possible by unchecking the DRM option from Amazon's Developer console, but now you can only do it from their SDK.

The `react-native-iap` module by default will enable Amazon's DRM, but if a user doesn't want to leverage it, there is no option to disable it.

With this PR a user can easily disable it by adding `isAmazonDrmEnabled=false` to their **gradle.properties** file.
The default will be `true`

Screenshot 2024-09-19 at 13 32 53

### DIFF
--- a/plugin/__tests__/fixtures/buildGradleFiles.ts
+++ b/plugin/__tests__/fixtures/buildGradleFiles.ts
@@ -20,7 +20,7 @@ android {
         versionCode 34
         versionName "1.16.2"
         buildConfigField "boolean", "IS_NEW_ARCHITECTURE_ENABLED", isNewArchitectureEnabled().toString()
-        buildConfigField "boolean", "ENABLE_AMAZON_DRM", getExtOrDefault("enableAmazonDrm")`;
+        buildConfigField "boolean", "IS_AMAZON_DRM_ENABLED", getExtOrDefault("isAmazonDrmEnabled")`;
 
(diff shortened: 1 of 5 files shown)


plugin/__tests__/fixtures/buildGradleFiles.ts
```diff
@@ -19,7 +19,8 @@ android {
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 34
         versionName "1.16.2"
-        buildConfigField "boolean", "IS_NEW_ARCHITECTURE_ENABLED", isNewArchitectureEnabled().toString()`;
+        buildConfigField "boolean", "IS_NEW_ARCHITECTURE_ENABLED", isNewArchitectureEnabled().toString()
+        buildConfigField "boolean", "ENABLE_AMAZON_DRM", getExtOrDefault("enableAmazonDrm")`;
```
hyochan, line 23: Could we make this look more like a boolean? Something like `isAmazonDrmEnabled` would be great.

hyochan (changes requested):

hyochan:
Additionally, providing documentation related to this feature would be great

hyochan (approved):
LGTM!!

PR #2874, opened 2024-10-31
fix: storekit2 add verificationResultIOS to getAvailablePurchases

charlyBerthet:
As mentioned here: https://github.com/hyochan/react-native-iap/issues/2533 `verificationResultIOS` is not present when we call `getAvailablePurchases`. Therefore we are not able to verify the transaction from the server side.

We just need to send the verification object to the already implemented serializer.


hyochan (approved):
LGTM!

PR #2875, opened 2024-11-01
refactor: make the code parameter of safeReject non-nullable

mlazari:
The reject method of Promise class became non-nullable in React Native 0.76 after rewriting it from Java to Kotlin. We fixed it being called with a nullable value in https://github.com/hyochan/react-native-iap/pull/2872 by providing and empty string as a fallback if it happens to be null. But after more investigation it seems like it's not called with a value that could be null anywhere in the code, so a better solution would be to just update the type of the `code` parameter to be non-nullable.

### DIFF
--- a/android/src/main/java/com/dooboolab/rniap/PromiseUtils.kt
+++ b/android/src/main/java/com/dooboolab/rniap/PromiseUtils.kt
@@ -33,7 +33,7 @@ object PromiseUtils {
 
     fun rejectPromisesForKey(
         key: String,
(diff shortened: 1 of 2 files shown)


hyochan (approved):
Thanks for the updates!  
It seems like the type mismatch issue was due to a Kotlin internal type system inference rather than an error with the React Native API itself. Changing the `code` parameter from `String?` to `String` effectively resolved the inferred type mismatch.

PR #2877, opened 2024-11-04
feat: getStorefront method implementation for fetching App Store and Google Play region

nahlebn1k:
Hello, I've been using this method in my fork for a while. And i think it would be very usefull to have this in main repo.


nahlebn1k:
@hyochan Hi! The purpose of this PR is to add the ability to determine which App Store or Play Store region user in. In my case i have different in-app purchases and subscriptions packs for different regions. And in order to offer users specific purchases it's necessary to know their Store region.

In addition Android getBillingConfigAsync() method should be used with existing BillingClient, so the implementation of this feature suggests itself. Thanks in advance!

hyochan (approved):
Thanks for the explanation!

LGTM!

hyochan:
Documentation could be included in the future though!

PR #2896, opened 2024-12-12
Remove useless React-Codegen pod dependency

adrianrdzv:
This dependency is not needed (empty dependency) and is affecting building tvos projects using react-native-iap:
```
[!] The platform of the target `project-tvOS` (tvOS 15.1) is not compatible with `React-Codegen (0.1.0)`, which does not support `tvOS`.
```


hyochan (approved):
LGTM!

PR #2904, opened 2024-12-25
fix: expo 52 config plugin

Ash-hs:


shubhamdeol (approved):

mirland (approved):
Fixes: https://github.com/hyochan/react-native-iap/issues/2910 https://github.com/hyochan/react-native-iap/issues/2812 

Android support library was deprecated in 2018. 

Doc: https://developer.android.com/topic/libraries/support-library

So the default implementation should use [AndroidX version](https://developer.android.com/jetpack/androidx).

It was causing some compilation errors in expo

hyochan:
@mirland @shubhamdeol  
Thank you so much for your review. I trust the reviewers and will incorporate this PR into the new version.

@Ash-hs  
Thank you for your contribution 👍 Next time, even a short description would be a great help 🙇🏻‍♂️

PR #2922, opened 2025-01-25
fix transactionListener jsdoc

vafada:
this fixes the `transactionListener` jsdoc

seems like the jsdoc `transactionListener` in `main` was copied and pasted from the jsdoc of `purchaseUpdatedListener`


hyochan:
Closes since this is not valid in version `14.+`! Thanks for the `PR` though

PR #2927, opened 2025-02-17
fix: add missing import in example

thecodeinfluencer:
- `finishTransaction` import was missing from the code snippet


hyochan (approved):
LGTM!

PR #2931, opened 2025-03-10
fix(ios): serialize UInt64 transaction IDs as strings to prevent E_DEVELOPER_ERROR during finishTransaction

Aflexon:
This pull request resolves a serialization issue affecting transaction identifiers (id and originalID) in the react-native-iap library for iOS. We've observed transaction IDs—particularly from family-shared subscriptions—with UInt64 values exceeding JavaScript’s MAX_SAFE_INTEGER (2^53 - 1), such as 504319184420155588.

When these large integer IDs are transmitted from native Swift code to JavaScript via the React Native bridge, they lose numeric precision. This loss causes transaction IDs to become corrupted, resulting in errors like E_DEVELOPER_ERROR with the message Invalid transaction Id when calling finishTransaction with IDs obtained from the purchaseUpdatedListener. Consequently, transactions may fail to complete properly.

**Proposed Changes**
1. Serialize transaction identifiers (id, originalID, webOrderLineItemID) explicitly as strings to maintain numerical precision.
2. Retain numeric serialization for timestamp fields (millisecondsSince1970), as they currently do not risk exceeding JavaScript's precision limit.


hyochan:
Closes since this is not valid in recent version starting from `v14.+`. Thanks for the `PR` though

PR #2953, opened 2025-05-29
feat(docs): add link to expo-iap in docs

Navipro70:
### **Description:**

I think we need add expo-iap to all documentation references, not only react-native-iap, but also to react-native-directories and expo itself, for community attention and better open-source maintaining 

### **Why I add this change:**

expo-iap is hard to find if you skip announcement of migration, but it can be very helpful for any developer, especially who want library with new architecture or expo support. Also expo-iap provide only StoreKit 2 and Google Play Billing, makes API and use cases more understandable

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,9 @@
 
 ## Announcement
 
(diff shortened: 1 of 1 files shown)


hyochan (approved):
Fabulous!

PR #2980, opened 2025-08-12
fix: unwrap optional status type in iOS StoreKit2 logic

elliotdickison:
This fixes the following Swift build error when attempting to build the package with Xcode 16.4 (16F6) for iOS 18.6 (22G86):

```
Value of optional type 'Product.SubscriptionInfo.Status?' must be unwrapped to refer to member 'renewalInfo' of wrapped base type 'Product.SubscriptionInfo.Status'
```

### DIFF
--- a/.github/workflows/ci-example-ios.yml
+++ b/.github/workflows/ci-example-ios.yml
@@ -43,7 +43,7 @@ jobs:
       - name: Select Xcode version
         uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd # v1.6.0
         with:
-          xcode-version: '16.3'
+          xcode-version: '16.4'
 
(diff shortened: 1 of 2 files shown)


elliotdickison:
I see that this is now breaking CI builds... should probably have opened an issue first. My change definitely fixes local builds using the versions I referenced. Seems that possibly Swift changed behavior here between Xcode 16.3 (CI) and Xcode 16.4 (my local version)?

elliotdickison:
Closing this as it appears to be a "me" issue. Unfortunately though I can't build the package as-is - my local machine simply disagrees with the CI machine on the presence of those optional chaining operators. I've tried compiling with Swift v4, v5 (used by CI), and v6 and I get the same error across the board. As we'll eventually move to `expo-iap` I'm simply going to patch the package locally and move on.

PR #3022, opened 2025-09-23
Add support for "auto-renewable-subscription" in iOS product type normalization

irekrog:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added support for the hyphenated iOS product type input "auto-renewable-subscription," ensuring it’s recognized alongside existing variants.
  * Prevents misclassification and related errors when configuring iOS subscriptions, improving reliability for different input formats.
  * Enhances compatibility without affecting existing setups or workflows.
  * Backwards compatible; no action required from users.


hyochan (approved):
LGTM!

PR #3034, opened 2025-09-26
Remove circular import between index & helper subscription

delphinebugner:
## Problem

Here's the circular import we have:

image

As a consequence in my PNPM project, when calling `hasActiveSubscriptions`, it failed with this error:

image

It could not resolve `getAvailablePurchases`. 

## Fix 

Easy fix is to move back the helpers in the `index.ts` file, worked on my project. 

Happy to try something else - but as `getAvailablePurchases` uses the `IAP` singleton of index.ts, i'm not sure of what to do :/

## Summary by CodeRabbit

- New Features
  - getActiveSubscriptions and hasActiveSubscriptions are now exposed on the main API.
  - checkTypeSynchronization is publicly available.
- Improvements
  - Backwards-compatible aliases added for acknowledgePurchase and consumePurchase.
- Refactor
  - Subscription logic moved from helper into the main public surface.
- Tests
  - Subscription-related tests reorganized and expanded; some helper-level tests removed and replaced with updated suites and per-test spies.

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -8,7 +8,7 @@ PODS:
   - hermes-engine (0.81.1):
     - hermes-engine/Pre-built (= 0.81.1)
   - hermes-engine/Pre-built (0.81.1)
-  - NitroIap (14.4.6):
+  - NitroIap (14.4.7):
     - boost
     - DoubleConversion
     - fast_float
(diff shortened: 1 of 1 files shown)


hyochan (approved):
Hello! Thank you for the PR 🙏
I've managed to fix rest of the part!

PR #3044, opened 2025-10-02
fix(android): prevent product data overwrite in fetchProducts with type 'all'

CacaoRick:
## Problem

When using `fetchProducts` with `type: 'all'` to query mixed product types (e.g., both InApp and Subscription SKUs), InApp products lose their price information.

### Reproduction Steps
1. Call `fetchProducts(['inapp_sku', 'subs_sku'], 'all')`
2. Check the returned InApp product
3. Observe that `price` field is `null` and `displayPrice`, `currency` fields are `""`

### Root Cause

The original implementation handled `ProductQueryType.All` by:
1. First querying InApp type → finds `inapp_sku` with complete price information
2. Then querying Subs type → queries with the same SKU, Google Play returns empty/incomplete data
3. Using `map[sku] = product` to unconditionally overwrite → valid InApp product data gets overwritten

```kotlin
// Original implementation
fetched.forEach { collected[it.id] = it }  // ❌ Unconditional overwrite
```

## Solution
OpenIAP's `fetchProducts` method already correctly implements `ProductQueryType.All` logic:
Uses processedIds to track already-processed product IDs
Prevents duplicate additions of the same product
Correctly handles mixed-type queries

Reference: [OpenIAP fetchProducts implementation](https://github.com/hyodotdev/openiap-google/blob/c0d224aebe34ee6baa1037fbef855cad6ae4781d/openiap/src/main/java/dev/hyo/openiap/OpenIapModule.kt#L149)

This PR removes the redundant manual implementation and directly uses OpenIAP's functionality.

## Testing

Tested in my app with the following scenarios:

- [x] Mixed InApp + Subs query (type: 'all') → Both product types now return complete information with correct pricing
- [x] InApp only query (type: 'inapp') → Works correctly
- [x] Subs only query (type: 'subs') → Works correctly

Before this fix, InApp products would return `{ displayPrice:"", currency:"", price:null }` when queried together with subscription products using type `'all'`. After this fix, all products maintain their complete data.

## Summary by CodeRabbit

- New Features
  - None

- Refactor
  - Streamlined in-app purchase product fetching into a single, consistent flow across product types.

- Performance
  - Reduced overhead by eliminating redundant aggregation, improving product load times.

- Stability
  - More consistent and reliable product retrieval for all product types.

- Observability
  - Simplified and consolidated logging for clearer diagnostics during product fetch operations.

### DIFF
--- a/android/src/main/java/com/margelo/nitro/iap/HybridRnIap.kt
+++ b/android/src/main/java/com/margelo/nitro/iap/HybridRnIap.kt
@@ -194,36 +194,15 @@ class HybridRnIap : HybridRnIapSpec() {
             val queryType = parseProductQueryType(type)
             val skusList = skus.toList()
 
-            val products: List<ProductCommon> = when (queryType) {
-                ProductQueryType.All -> {
-                    val collected = linkedMapOf<String, ProductCommon>()
-                    listOf(ProductQueryType.InApp, ProductQueryType.Subs).forEach { kind ->
-                        RnIapLog.payload(
-                            "fetchProducts.native",
-                            mapOf("skus" to skusList, "type" to kind.rawValue)
-                        )
-                        val fetched = openIap.fetchProducts(ProductRequest(skusList, kind)).productsOrEmpty()
-                        RnIapLog.result(
-                            "fetchProducts.native",
-                            fetched.map { mapOf("id" to it.id, "type" to it.type.rawValue) }
-                        )
-                        fetched.forEach { collected[it.id] = it }
-                    }
-                    collected.values.toList()
-                }
-                else -> {
-                    RnIapLog.payload(
-                        "fetchProducts.native",
-                        mapOf("skus" to skusList, "type" to queryType.rawValue)
-                    )
-                    val fetched = openIap.fetchProducts(ProductRequest(skusList, queryType)).productsOrEmpty()
-                    RnIapLog.result(
-                        "fetchProducts.native",
-                        fetched.map { mapOf("id" to it.id, "type" to it.type.rawValue) }
-                    )
-                    fetched
-                }
-            }
+            RnIapLog.payload(
+                "fetchProducts.native",
+                mapOf("skus" to skusList, "type" to queryType.rawValue)
+            )
+            val products = openIap.fetchProducts(ProductRequest(skusList, queryType)).productsOrEmpty()
+            RnIapLog.result(
+                "fetchProducts.native",
+                products.map { mapOf("id" to it.id, "type" to it.type.rawValue) }
+            )
 
             products.forEach { p -> productTypeBySku[p.id] = p.type.rawValue }
 


CacaoRick:
Wait, I noticed that there is a bug in the cache of OpenIapModule's `fetchProducts` method for `ProductQueryType.All`. It may need to be handled differently.

The first `fetchProducts` will get the correct result, but after the second time, the type of `subs` also becomes `in-app`.
```
fetchProducts payload: {"skus":["subs_sku","inapp_sku"],"type":"all"}
initConnection payload: null
initConnection result: true
fetchProducts.native payload: {"skus":["subs_sku","inapp_sku"],"type":"all"}
fetchProducts.native result: [{"id":"subs_sku","type":"subs"},{"id":"inapp_sku","type":"in-app"}]
fetchProducts result: [{"id":"subs_sku","type":"subs"},{"id":"inapp_sku","type":"in-app"}]

fetchProducts payload: {"skus":["subs_sku","inapp_sku"],"type":"all"}
initConnection payload: null
initConnection result: true
fetchProducts.native payload: {"skus":["subs_sku","inapp_sku"],"type":"all"}
fetchProducts.native result: [{"id":"subs_sku","type":"in-app"},{"id":"inapp_sku","type":"in-app"}]
fetchProducts result: [{"id":"subs_sku","type":"in-app"},{"id":"inapp_sku","type":"in-app"}]

```

final line `{"id":"subs_sku","type":"in-app"}` type is wrong

CacaoRick:
## Closing This PR

After further investigation, I've discovered that this issue stems from bugs in the underlying OpenIAP library that cannot be reliably worked around at the react-native-iap level.

### Root Cause Analysis

#### Issue 1: OpenIAP `ProductQueryType.All` Cache Bug

OpenIAP's `fetchProducts` with `ProductQueryType.All` appears to have a caching issue that causes incorrect results on subsequent queries.

#### Issue 2: Type Mismatch Returns Incorrect Product Types

When querying with `ProductQueryType.Subs` or `ProductQueryType.InApp`, OpenIAP returns **all provided SKUs marked with the queried type**, even if they don't actually match that type:

```kotlin
// Query with Subs type but include an InApp SKU
fetchProducts(["subs_sku", "inapp_sku"], ProductQueryType.Subs)

// OpenIAP returns BOTH marked as "subs"
Result: [
  {"id": "subs_sku", "type": "subs"},     // ✅ Correct
  {"id": "inapp_sku", "type": "subs"}     // ❌ Wrong! Actually an in-app product
]
```

**This PR will remain closed until**
- OpenIAP fixes the ProductQueryType.All caching bug
- OpenIAP properly handles type validation (only returning products that match the requested type)

Once these issues are resolved in OpenIAP, this PR approach (or a similar fix) can be reconsidered.

**Workaround for Users**
Until OpenIAP is fixed, users experiencing this issue should query product types separately and ensure each SKU is queried with its correct type: 

❌ Don't use:

```javascript
await fetchProducts(['inapp_sku', 'subs_sku'], 'all')
```

✅ Instead, query with correct types:

```javascript
const inappProducts = await fetchProducts(['inapp_sku'], 'inapp')
const subsProducts = await fetchProducts(['subs_sku'], 'subs')
const allProducts = [...inappProducts, ...subsProducts]
```

**Related Issues**
This issue should be reported to the OpenIAP repository:
- OpenIAP ProductQueryType.All caching issue
- Type mismatch returning incorrect product types

---
Closing this PR. Can be reopened once OpenIAP resolves the underlying issues.

hyochan:
@CacaoRick Hey good catch! I'll manage to fix this here https://github.com/hyodotdev/openiap-apple/pull/16 and release to `1.2.6`. Hope this works.

hyochan:
@CacaoRick Released `14.4.10`

CacaoRick:
Thank you for your maintenance.

